PR #5, opened 2015-10-29
Fixed issue where if delegate was called without being set caused a

gus-byrnesinnovation:
Updated Podspec properly

Error was that it was trying to call the method on a class not the
instance.

Added additional safety net just in case the delegate gets removed before the callback returns and we could possible get a nil error exception,


jkodumal:
:+1: 

Pointed `hello-ios` to `cm-ldclient-delegate` and all seems to work.

PR #7, opened 2015-10-30
Changed so that all operations touching the managedObjectContext on the

gus-byrnesinnovation:
background thread use the contexts serial queue to execute.

Taken from:

http://stackoverflow.com/questions/9922145/what-is-nsmanagedobjectcontexts-performblock-used-for


jkodumal:
FYI, it's probably best *not* to create a tag corresponding to a branch SHA. The SHA is going to change when this gets merged to master, and that's the SHA that should be tagged.

PR #13, opened 2015-11-19
Event capacity

### LINKED ISSUE #12: Capacity for Events
@jkodumal 

In working on the configuration issue, @gus-byrnesinnovation and I noticed that we are not using the capacity value that can be passed in via configBuilder on starting the client. We wanted to confirm what this capacity means. Our interpretation is that when events are persisted we confirm that the collection of events is less than this value. We then disregard any events that occur after. When, the next flush happens, the persisted events will be cleared out, and then we will begin persisting again. Is this understanding correct?

cmavromoustakos:
Fixes #12

https://github.com/launchdarkly/ios-client/issues/12

### DIFF
--- a/Darkly/ClientManager.m
+++ b/Darkly/ClientManager.m
@@ -18,8 +18,7 @@
 
 NSString *const kLDUserUpdatedNotification = @"Darkly.UserUpdatedNotification";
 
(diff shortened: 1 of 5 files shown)


jeff-byrnesinnovation:
Looks good to me. Merging

PR #20, opened 2015-12-04
Changed so that we set the properties directly on the entity. Mantle

cmavromoustakos:
does not save associations to core data.

Added check if config exists and if it does merge the dictionary values
from the server response and save the context

Using uniquing key and generating key on config from user key.

We are forcing mantle to update the config entry based ont he user key
attached.

created migration for new field

explicitly set userKey. We will get a recursive loop if we try to set
user and config/

fixes https://github.com/launchdarkly/ios-client/issues/19


Darkly/ClientManager.m
```diff
@@ -131,23 +131,26 @@ - (void)processedEvents:(BOOL)success jsonEventArray:(NSData *)jsonEventArray ev
...
-        DEBUG_LOGX(@"ClientManager processedConfig method called after receiving successful response from server");
+           DEBUG_LOGX(@"ClientManager processedConfig method called after receiving successful response from server");
         // If Success
-        LDFlagConfig *config = [MTLJSONAdapter modelOfClass:[LDFlagConfig class]
-                                   fromJSONDictionary:jsonConfigDictionary
-                                                error: nil];
-        if (config) {
+        LDFlagConfig *newConfig = [MTLJSONAdapter modelOfClass:[LDFlagConfig class]
+                                            fromJSONDictionary:jsonConfigDictionary
+                                                         error: nil];
```
cmavromoustakos: user.config = newConfig

Darkly/ClientManager.m
```diff
@@ -131,23 +131,27 @@ - (void)processedEvents:(BOOL)success jsonEventArray:(NSData *)jsonEventArray ev
 
 - (void)processedConfig:(BOOL)success jsonConfigDictionary:(NSDictionary *)jsonConfigDictionary configInterval:(int)configInterval {
     if (success) {
-        DEBUG_LOGX(@"ClientManager processedConfig method called after receiving successful response from server");
+           DEBUG_LOGX(@"ClientManager processedConfig method called after receiving successful response from server");
```
jeff-byrnesinnovation: Do we need this indent?

Darkly/ClientManager.m
```diff
@@ -155,8 +159,7 @@ - (void)processedConfig:(BOOL)success jsonConfigDictionary:(NSDictionary *)jsonC
             
             [[NSNotificationCenter defaultCenter] postNotificationName: kLDUserUpdatedNotification
                                                                 object: nil];
-        }
-    } else {
+        }} else {
```
jeff-byrnesinnovation: Can we fix the indentation for this?

Darkly/Models/DataManager.m
```diff
@@ -66,11 +66,14 @@ -(void) saveUser: (LDUser *) user {
...
-                               insertingIntoContext: [[DataManager sharedManager] managedObjectContext]
-                                              error: nil];
     
-    [[DataManager sharedManager] saveContext];
+    [self.managedObjectContext performBlockAndWait:^{
+        [MTLManagedObjectAdapter managedObjectFromModel: user
+                                   insertingIntoContext: [[DataManager sharedManager] managedObjectContext]
+                                                  error: nil];
+        
+        [[DataManager sharedManager] saveContext];
```
jeff-byrnesinnovation: Should this be [self saveContext]?

Darkly/Models/DataManager.m
```diff
@@ -66,11 +66,14 @@ -(void) saveUser: (LDUser *) user {
         [[DataManager sharedManager] purgeOldUsers];
```
jeff-byrnesinnovation: In DataManager, sometimes you use [DataManager sharedManager] instead of self. Also, in at least one method, DataManager.purgeOldUsers, you are saving the context outside of the block.

cmavromoustakos: Nice catch.

jeff-byrnesinnovation:
Looks good to me.

PR #37, opened 2016-02-24
Fixed issue with date formatter

cmavromoustakos:
added test for dictionary value for user model

### DIFF
--- a/Darkly/LDUserModel.m
+++ b/Darkly/LDUserModel.m
@@ -31,7 +31,8 @@ static NSString * const kOsKey = @"os";
     NSMutableDictionary *dictionary = [[NSMutableDictionary alloc] init];
     
     NSDateFormatter *formatter = [[NSDateFormatter alloc] init];
-    [formatter setDateFormat:@"yyyy-MM-ddTHH.mm.ss"];
+    [formatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"];
+    [formatter setTimeZone:[NSTimeZone timeZoneWithAbbreviation:@"UTC"]];
     
     self.key ? [dictionary setObject:self.key forKey: kKeyKey] : nil;
     self.ip ? [dictionary setObject:self.ip forKey: kIpKey] : nil;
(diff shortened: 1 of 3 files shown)


Darkly/LDUserModel.m
```diff
@@ -101,9 +101,7 @@ - (id)initWithDictionary:(NSDictionary *)dictionary {
         self.device = [dictionary objectForKey: kDeviceKey];
         self.os = [dictionary objectForKey: kOsKey];
         self.anonymous = [dictionary objectForKey: kAnonymousKey];
-        if (self.config) {
-            self.config = [[LDFlagConfigModel alloc] initWithDictionary:[dictionary objectForKey:kConfigKey]];
-        }
+        self.config = [[LDFlagConfigModel alloc] initWithDictionary:[dictionary objectForKey:kConfigKey]];
         if (self.updatedAt) {
```
jeff-byrnesinnovation: I guess we can remove updatedAt here since it will be created by [self init]

Darkly/LDUserModel.m
```diff
@@ -31,7 +31,7 @@ -(NSDictionary *)dictionaryValue{
     NSMutableDictionary *dictionary = [[NSMutableDictionary alloc] init];
     
     NSDateFormatter *formatter = [[NSDateFormatter alloc] init];
-    [formatter setDateFormat:@"yyyy-MM-ddTHH.mm.ss"];
+    [formatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"];
```
jeff-byrnesinnovation: how come we have to change this? I.e. I think i pulled this from the old code, was this not correct?

DarklyTests/Models/LDUserModelTest.m
```diff
@@ -26,4 +26,64 @@ -(void)testNewUserSetupProperly {
     XCTAssertNotNil(user.updatedAt);
 }
 
+-(void)testDictionaryValue {
```
jeff-byrnesinnovation: For this test do we not simply want to create a user, call dictionaryValue on it, then initWithDictionary on it? Then compare the initial and recreated objects?

PR #38, opened 2016-02-24
Removed unecesary delegate propery definition within the protocol

cmavromoustakos:
itself.


Darkly/LDClient.h
```diff
@@ -19,7 +18,7 @@
 
 @property(nonatomic, strong, readonly) LDUser *user;
 @property(nonatomic, strong, readonly) LDConfig *ldConfig;
-@property (nonatomic, assign) id delegate;
+@property (nonatomic, assign) id<ClientDelegate> delegate;
```
jeff-byrnesinnovation: I think you want weak here right?

PR #50, opened 2016-05-21
Remove timer nullifying

danialzahid94:


jkodumal:
👍 This should also target master as well, correct?

drichelson:
I cherry-picked this onto master and it is in the 0.4.4 release as well.

PR #55, opened 2016-07-10
V2 Implementation

danialzahid94:
This PR includes:

- Functionality for multivariate flags
- Real time Event based polling
- Name base convention upgrade
- Support for new v2 backend structure


Darkly/LDClient.h
```diff
@@ -39,7 +40,27 @@
...
+/**
+ * Retrieve a feature flag value. If the configuration for this feature
+ * flag is retrieved from the server that value is returned, otherwise
+ * the defaultValue is returned.
+ *
+ * @param featureName   Name of feature flag
+ * @param default   Default value for feature flag
+ * @return the feature flag value
+ */
+- (NSString*)stringVariation:(NSString *)featureKey defaultValue:(NSString*)defaultValue;
```
jkodumal: In Java, we also have a `jsonVariation` method, as LD now supports json objects and arrays:

https://github.com/launchdarkly/java-client/blob/v2/src/main/java/com/launchdarkly/client/LDClient.java#L288

What's the objective C equivalent? Two methods that return `NSMutableDictionary` (for json objects) or `NSMutableArray` (for json arrays)? Or is there a common AST type we can return as in Java?

Darkly/LDConfig.h
```diff
@@ -2,6 +2,8 @@
 //  Copyright © 2015 Catamorphic Co. All rights reserved.
 //
 
+@class LDUserModel;
```
jkodumal: What is this change for?

Darkly/LDClient.h
danialzahid94: If the server is sending pure json strings, then yes, `NSDictionary` and `NSArray` will cover the two types properly. I'll make two separate functions for each of these.

Darkly/LDConfig.h
danialzahid94: I shifted the LDUserModel.h import from .m to .h file to provide the public `withUserModel:` method. The import was creating a cyclic redundancy, so added the class reference instead of importing `LDUserModel.h`.

Darkly/EventSource/EventSource.h
```diff
@@ -0,0 +1,100 @@
+//
```
jkodumal: Why is this inlined rather than specified as a CocoaPod dependency?

https://cocoapods.org/pods/EventSource

Darkly/LDClient.m
```diff
@@ -53,6 +55,13 @@ - (BOOL)start:(LDConfigBuilder *)inputConfigBuilder userBuilder:(LDUserBuilder *
                 LDClientManager *clientManager = [LDClientManager sharedInstance];
                 [clientManager syncWithServerForConfig];
                 [clientManager startPolling];
+                
+                EventSource *source = [EventSource eventSourceWithURL:[NSURL URLWithString:kStreamUrl] apiKey:ldConfig.apiKey];
```
jkodumal: Does this disconnect the stream when the app is backgrounded?

jkodumal: When backgrounding is set up, will we need to specify that we're doing background execution?

danialzahid94: Currently, this will keep listening to events even in background. In the next few commits I'll implement disconnecting the stream, and add hard polling after every 5 minutes.

Darkly/EventSource/EventSource.h
danialzahid94: We had to make changes to the library to support custom headers for Authentication, and if we use the pod we would have to replace those changes each time.

jkodumal: In that case we should fork the library and release our own pod.

Darkly/LDClientManager.m
```diff
@@ -26,6 +27,7 @@ +(LDClientManager *)sharedInstance {
         
         [[NSNotificationCenter defaultCenter] addObserver:sharedApiManager selector:@selector(willEnterForeground) name:UIApplicationWillEnterForegroundNotification object:nil];
         [[NSNotificationCenter defaultCenter] addObserver:sharedApiManager selector:@selector(willEnterBackground) name:UIApplicationWillResignActiveNotification object:nil];
+        [[NSNotificationCenter defaultCenter] addObserver:sharedApiManager selector:@selector(syncWithServerForEvents) name:kLDBackgroundFetchInitiated object:nil];
```
danialzahid94: @jkodumal Developers would need to add the `application:performFetchWithCompletionHandler:` to their AppDelegate and post this notification from there to make the background fetching work. Or if you have a more efficient idea? Because this method does not throw it's own notification.

jkodumal: Can you give me a complete example of how a developer would set this up?

danialzahid94: They will have to import LDClient.h and add this code to their AppDelegate:
````swift
func application(application: UIApplication, performFetchWithCompletionHandler completionHandler: (UIBackgroundFetchResult) -> Void) {
    NSNotificationCenter.defaultCenter().postNotificationName(kLDBackgroundFetchInitiated, object: nil)
}

Comment (jkodumal) [inline Darkly/LDClientManager.m]: Seems pretty good. Can you think of any other options?

Comment (danialzahid94) [inline Darkly/LDClientManager.m]: I think this is the most minimalistic approach we can provide, and given a proper documentation, it will be quite easy for the developers to integrate.

**Comment by danialzahid94** [comment]
@jkodumal Can you please review the latest changes before I merge the PR.

File: Darkly/DarklyConstants.m [hunk 6]
```diff
@@ -5,7 +5,7 @@
 #import "DarklyConstants.h"
 
 NSString * const kClientVersion = @"0.4.2";
-NSString * const kBaseUrl = @"https://app.launchdarkly.com/msdk/eval/users";
+NSString * const kBaseUrl = @"https://mobile.launchdarkly.com";
```
jkodumal: this won't work, will it? It changes the URL not just for events but for fetching flags.

Darkly/DarklyConstants.m
```diff
@@ -0,0 +1,28 @@
+//
+//  Copyright © 2015 Catamorphic Co. All rights reserved.
+//
+
+#import "DarklyConstants.h"
+
+NSString * const kClientVersion = @"0.4.2";
+NSString * const kBaseUrl = @"https://mobile.launchdarkly.com";
```
jkodumal: Mentioned in another comment that seems to have been marked outdated, but I don't think this will work. Only the events url should change to mobile-- this will change the feature flag polling as well.

danialzahid94:
@jkodumal Ready for a final review before I merge.

jkodumal:
👍

PR #63, opened 2016-08-17
Fix config flags in user model

danialzahid94:
@jkodumal Could you please review the changes before I merge.


pkaeding:
It looks like there is a test that needs to be modified now that we are not sending the config object in the event: https://circleci.com/gh/launchdarkly/ios-client/72 (unless the CircleCI bug with exit code 65 is causing this failure, but it seems like there is a real test failure, and it makes sense that changing the data in the event would need a test case change too.)

danialzahid94:
Just needed to restart the tests once. It was an internal CircleCI issue.

pkaeding:
lgtm
:+1:
looks good to me
merge away
I approve
canned replies are great.

PR #73, opened 2016-11-29
Remove AFNetworking dependency

danialzahid94:
This PR aims to remove the AFNetworking dependency from the SDK. Instead of AFNetworking, the SDK now uses native NSURLSession to make the web calls.


danialzahid94:
@jkodumal Up for final review before we merge it into master and release.

PR #79, opened 2017-01-04
Begin work on adding cross platform support

kdawgwilk:
This is the beginnings to add cross platform support.  We recently started using LaunchDarkly and built a small wrapper around the client but then realized it only supported iOS so I figured I would give it a shot to support other Apple platforms


kdawgwilk:
This is not ready to merge currently but I thought it might be good to get eyes on it asap

ebriney:
@kdawgwilk : When do you think it will be ready? Because I have the same problem...

arun251:
Hi @ebriney, thanks for the feedback. We will be updating and merging this PR this week.

Darkly/LDUserBuilder.m
```diff
@@ -191,7 +191,11 @@ -(LDUserModel *)build {
         [user key:key];
         DEBUG_LOG(@"LDUserBuilder building User with key: %@", key);
     } else {
+#if TARGET_OS_OSX
+        NSString *uniqueKey = @"";
```
kdawgwilk: Not sure how to get a unique key on macOS

Podfile
```diff
@@ -2,7 +2,7 @@
 # platform :ios, '6.0'
 
 target 'Darkly' do
-  pod 'DarklyEventSource', :git => 'https://github.com/launchdarkly/ios-eventsource.git'
+    pod 'DarklyEventSource', :git => 'https://github.com/kdawgwilk/ios-eventsource.git', :branch => 'feature/add-support-for-tvos-watchos'
```
kdawgwilk: This will need to be updated when the PR merges fixing the dependency

Pods/DarklyEventSource/EventSource/EventSource.m
```diff
@@ -203,17 +186,17 @@ - (void)connection:(NSURLConnection *)connection didReceiveData:(NSData *)data
...
 
     if (wasClosed) {
         return;
     }
-    
+
     Event *e = [Event new];
     e.readyState = kEventStateClosed;
-    e.error = [NSError errorWithDomain:@""
+    e.error = error ?: [NSError errorWithDomain:@""
```
kdawgwilk: Whoops not sure what happened here

arun251:
Thanks for the PR @kdawgwilk . We have considerable work on another branch that we intend to merge soon. We will focus on that branch instead.

PR #81, opened 2017-01-23
Automate builds using CI

danialzahid94:


circle.yml
```diff
@@ -4,4 +4,5 @@ machine:
 
 test:
  post:
-  - pod lib lint
\ No newline at end of file
+  - pod lib lint
+  - if [ "$CIRCLE_BRANCH" = 'master' ]; then pod spec lint; fi
```
jkodumal: Why only lint on Master?

danialzahid94: 'pod lib lint' only validates the pod locally, and does not check for the rights tags etc. Although, having this for any branch/PR is beneficial to verify and identify any issues before hand and fix within the same branch/PR.
'pod spec lint' does a full remote validation, and requires for the respective tag for tat version to be available on the master branch. Because of this, spec lint should be run when everything is finalized.
For this reason, pod spec lint is configured to run only when committed to master.

jkodumal:
👍

Issue #82, opened 2017-01-24
Terminating app due to uncaught exception 'NSInvalidArgumentException'


jimmaye:
Hi

Sometimes when running the client in the iOS simulator (iPhone 6s, iOS10.2), I get this crash - usually during launch of the application (build+run from Xcode):
```
2017-01-24 11:56:34.263 app-name[30637:2191253] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid type in JSON write (UIViewAnimationBlockDelegate)'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001126f9d4b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010cca421e objc_exception_throw + 48
	2   CoreFoundation                      0x00000001127632b5 +[NSException raise:format:] + 197
	3   Foundation                          0x000000010e2fc964 _writeJSONValue + 668
	4   Foundation                          0x000000010e352232 ___writeJSONObject_block_invoke + 226
	5   CoreFoundation                      0x0000000112680dc6 __65-[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]_block_invoke + 102
	6   CoreFoundation                      0x0000000112680cca -[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:] + 202
	7   Foundation                          0x000000010e352026 _writeJSONObject + 430
	8   Foundation                          0x000000010e2fc84d _writeJSONValue + 389
	9   Foundation                          0x000000010e2fcdb9 ___writeJSONArray_block_invoke + 132
	10  CoreFoundation                      0x00000001126890b2 __53-[__NSArrayM enumerateObjectsWithOptions:usingBlock:]_block_invoke + 114
	11  CoreFoundation                      0x0000000112688f6f -[__NSArrayM enumerateObjectsWithOptions:usingBlock:] + 239
	12  Foundation                          0x000000010e2fcabf _writeJSONArray + 330
	13  Foundation                          0x000000010e2fc8e4 _writeJSONValue + 540
	14  Foundation                          0x000000010e2fc674 -[_NSJSONWriter dataWithRootObject:options:error:] + 124
	15  Foundation                          0x000000010e2fc554 +[NSJSONSerialization dataWithJSONObject:options:error:] + 333
	16  LaunchDarkly                        0x000000010e82b341 -[LDRequestManager performEventRequest:] + 405
[...]

PR #83, opened 2017-01-24
Fix for crash reported in #82 

jimmaye:
Changes 'value' and 'isDefault' properties to be strongly retained to avoid crashing when having feature toggles that's of the multivariation type. Crash is mitigated when only true/false feature flags is used - which makes sense when going through the code.


arun251:
Thanks for the fix, @jimmaye . We will release an update soon.

arun251:
This was just released in version 2.1.3.

PR #89, opened 2017-04-26
Updates to LDConfigBuilder API

petrucci34:
### Release Notes
* This PR attempts to provide a new look at the two core objects, `LDConfig` and `LDConfigBuilder` and suggests some fundamental architectural changes.
* `LDConfig` is an object that stores key information such as the mobile key, base and event URLs, flush/poling time intervals etc. `LDConfigBuilder` solely exists to allocate and initialize an `LDConfig` object. In doing so, `LDConfigBuilder` redundantly keeps a copy of every single `LDConfig` property. It then creates a config object and returns a reference to it.
* This PR gets rid of `LDConfigBuilder` and instead builds necessary mechanisms into `LDConfig` object so that:
    - Required and optional properties are managed through the compiler via `nonnull` and `nullable` attributes.
    - It's not possible to create a config object without specifying a mobile key.
    - There are no redundant temporary variables.
* The PR also introduces Objective-C 2.0 syntax for properties as opposed to the now obsolete instance-variable declaration, getter/setter synthesis pattern.
* Made sure the tests pass and all references to the builder objects are replaced with direct instantiation of config objects.


arun251:
Thanks, @petrucci34 . We'll review this PR soon.

arun251:
Thanks again for the PR, @petrucci34 !

PR #91, opened 2017-05-18
Fix semaphore issues

arun251:
This PR intends to fix the semaphore bug which was resulting in multiple api calls being generated simultaneously after the time specified in `kConnectionTimeout`.


arun251:
👍

PR #93, opened 2017-05-19
Optimize persistence of feature flags

arun251:
This PR optimizes user persistence by removing encoding and decoding of NSData each time a user has to be updated. The performance improvement of this change is significant and allows for multiple threads to perform users creations without blocking other activities on the main thread.


arun251:
Thanks, @danialzahid94 !

Issue #95, opened 2017-05-25
Client crashes if a feature is off and has no off-variation


arun251:
Earlier versions of the iOS SDK can crash if the feature flag is in a particular state:
* the feature flag targeting is off
* there is no default variation specified when the flag's targeting is off: ![no-off-variation](https://cloud.githubusercontent.com/assets/1594156/26468767/719c1ac0-414c-11e7-9a75-470ba57fc93b.png)

Upgrading to the latest release (2.3.3 or later) is encouraged. As a workaround, you can set the "off variation" for all feature flags where targeting is off.

arun251:
Fixed in version 2.3.3.

PR #96, opened 2017-05-25
Fix flag storage

arun251:
This PR is to fix the flag storage crash (#95) if null is returned from the server for a flag value.


arun251:
This looks good. Thanks for the fix, @danialzahid94 .

Issue #97, opened 2017-06-01
Memory leak in LDRequestManager.m (with regards to NSURLSession)


jimmaye:
Hi

First of all, thanks once more for doing this service, and keeping the code open, it's really a pleasure and lifesaver for us.

While profiling our app, I discovered quite an amount of leaks (even when our app where idling), and I found. the following stack trace:
![ldrequestmanager-leak](https://cloud.githubusercontent.com/assets/6573169/26677131/e055458a-46ca-11e7-9bf0-be5f5f6278f7.png)

I'll submit a PR which changes the usage of `NSURLSession` from `[NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration]]` to `[NSURLSession sharedSession]`. This will effectively remove those leaks. Another option could be to use the ephemeral `[NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration ephemeralSessionConfiguration]]` configuration instead - but since we have no need for `NSURLSession`'s delegate callbacks, the shared session might be enough. You might have better insights into if some of the side effect of using the shared session is unwanted.

I hope you choose to include this in a forthcoming SDK release any time possible :-)

jimmaye:
PR here: https://github.com/launchdarkly/ios-client/pull/98

arun251:
Thanks for reporting and submitting PRs!

arun251:
This was fixed in release 2.4.0. Thanks again.

PR #98, opened 2017-06-01
#97: Addressing memory leak in LDRequestManager

jimmaye:
This pull requests deals with the `NSURLSession` memory leaks in the `LDRequestManager`.


arun251:
Thanks for the patch @jimmaye . We will release a new version tomorrow.

PR #100, opened 2017-06-20
Propagate the SDK version to ios-eventsource as User-Agent

arun251:
This PR intends to reflect changes to ios-eventsource and pass user agent along with mobile key to the event source.


jkodumal:
👍

PR #101, opened 2017-06-21
Fix race conditions around polling states

arun251:
This PR fixes the data races reported by the thread sanitizer, and synchronizes the read/write operations for the polling states.


arun251:
This looks good to me, thanks!

PR #105, opened 2017-07-07
Add name in LDUserModel

arun251:
This PR intends to add a name property to `LDUserModel` and its respective builder methods to the `LDUserBuilder` class.


arun251:
Good idea. Done.

PR #106, opened 2017-07-13
Resolve bugs in the iOS client

atlassian-gaustin:
This brings the iOS client up to full feature parity


arun251:
Hi @atlassian-gaustin thanks for the PR! Do you mind describing the motivations behind these commits?

atlassian-gaustin:
The iOS client was badly broken — example: you can’t just typecast a string into a dictionary or array.

Also, the long-lived network connection would time out and close because the server wouldn’t have any changes to push.

atlassian-gaustin:
Also, there is no reason to keep around sources retrieved via CocoaPods (or Carthage). Let CocoaPods or Carthage manage the sources.

atlassian-gaustin:
Incorporate necessary changes into roll-up PR.

PR #107, opened 2017-07-21
Allow notification of the client if the server connection to retrieve flags fails.

atlassian-gaustin:
If the connection to retrieve the current flag state fails, clients _may_ want to know about that failure.
This is especially true if the flags provide required information not available at build time.


atlassian-gaustin:
Incorporated into roll-up PR

PR #112, opened 2017-08-31
Add the ability to be notified if a server update resulted in no change.

atlassian-gaustin:
This is far more critical in polling mode versus streaming mode, but it is useful to know if the values that we currently have (or can get from the client) are current.


arun251:
@atlassian-gaustin I'm curious about the use case for this. I would expect this notification to fire very frequently in polling mode and fairly often in streaming mode.

atlassian-gaustin:
Yes, it will fire each time in polling mode if there are no changes. That is deliberate for the case where you need to make sure that a connection was made, was successful, yet there were no changes. It's entirely optional whether you use that information but it orthogonal to tracking changes and failures to connect to the server.

arun251:
This is released in version 2.9.0. Thanks again!

PR #113, opened 2017-09-11
Changes to make the Darkly framework extension-safe.

atlassian-gaustin:
There's no need for the Client Manager to be a UIApplicationDelegate, and the resulting framework needs to be marked as extension-safe.


arun251:
This is released in version 2.9.0. Thanks again!

PR #123, opened 2017-12-05
updates to version 2.9.1

markpokornycos:
Fixes a Carthage build issue that could cause the sdk to crash at runtime. A corresponding CocoaPods change keeps the version numbers up-to-date.

## Testing steps
In the hello-tvOS project directory
1. Delete `hello-tvOS/Carthage`
2. edit `Cartfile` replacing the existing git reference with `git "file:///Path/to/ios-client" "version_2.9.1"
` (Note the 3 `/` in the path)
3. from the command line in the hello-tvOS project directory, run `carthage update`. Verify the build succeeds. 
4. Open `hello-tvOS/Carthage/Build`. Verify each OS has a newly built `Darkly.framework` and `DarklyEventSource.framework`
5. Open hello-tvOS.xcodeproj.
6. Run on any of the tv simulators. Verify the app does not crash.


Pods/Local Podspecs/DarklyEventSource.podspec.json
```diff
@@ -9,15 +9,15 @@
   },
   "source": {
     "git": "https://github.com/launchdarkly/ios-eventsource.git",
-    "tag": "3.1.0"
+    "tag": "v3.1.1"
```
arun251: This is probably not right -- we don't prefix tags with `v`.

markpokornycos: I saw this too, but it's an artifact of CocoaPods. The LDEventSource.podspec doesn't have the 'v' in the tag.

I tried pod install using 3.1.1, followed by pod update to return to 3.1.2. That didn't change this.

If you want me to track this down I will, but I suspect it came from releasing 3.1.1 with the `v` in the tag.

PR #124, opened 2018-01-08
Add additional check to comparison of user config flag values

natechiger:
I came across this issue when starting LaunchDarkly as an anonymous user.

Upon the first installation of the app, the user is signed out and registered as anonymous. The dictionary that is saved to `NSUserDefaults` does not contain any config information as it has not been fetched yet.
```
{
    anonymous = 1;
    key = "63EC2339...";
    ..
}
```

Once the config is fetched, it's saved on the registered user.
```
{
    anonymous = 1;
    config =     {
        "my-custom-flag" = 0;
    };
    key = "63EC2339...";
    ..
}
```

However, the comparison being done in `compareConfigForUser:withNewUser:` does not account for the config not existing at all. So in the scenario above, this comparison `![[newUser.config configFlagValue:key] isEqual:userVal]` will always be true which results in the notification being posted even though the flag did not change.

Adding a check to make sure the `user` config value is not nil before posting the notification solved my use case. But if this is done by design or there is a better way around this, please let me know.

### DIFF
--- a/Darkly/LDDataManager.m
+++ b/Darkly/LDDataManager.m
@@ -99,7 +99,8 @@ dispatch_queue_t eventsQueue;
 
 - (void)compareConfigForUser:(LDUserModel *)user withNewUser:(LDUserModel *)newUser {
     for (NSString *key in [[newUser.config dictionaryValue] objectForKey:kFeaturesJsonDictionaryKey]) {
-        if(user == nil || ![[newUser.config configFlagValue:key] isEqual:[user.config configFlagValue:key]]) {
+        NSObject *userVal = [user.config configFlagValue:key];
+        if(user == nil || (userVal != nil && ![[newUser.config configFlagValue:key] isEqual:userVal])) {
             [[NSNotificationCenter defaultCenter] postNotificationName:kLDFlagConfigChangedNotification object:nil userInfo:[NSDictionary dictionaryWithObject:key forKey:kFlagKey]];
         }
     }


markpokornycos:
[Review GH Issue #124](https://trello.com/c/OA56SkrG/100-review-gh-issue-124)

markpokornycos:
@natechiger Thank you for your PR! We appreciate you taking the time to explain the issue and recommend a solution.
From the description, it appears you are seeing the sdk post the notification `kLDFlagConfigChangedNotification`, and call the LDClient delegate method `featureFlagDidUpdate` when the sdk receives the first feature flag update for a new user. The user being anonymous or identified shouldn’t affect the sdk’s behavior in this particular case, but it is relevant that the sdk has not seen the particular user before.

In the sample above, the server returns `0` for `my-custom-flag`. Prior to receiving the flag config, the sdk will report the fallback value provided in the `variation` method called. When the sdk receives the flag config from the server, the sdk will register the new value of `0` as a change because the previous value was not defined. If the fallback value was `0` in the original `variation` request, then posting the `kLDFlagConfigChangedNotification` and calling the `featureFlagDidUpdate` method might seem superfluous. The sdk does not retain fallback values, and so when the sdk processes the flag config, it does not “remember” the last fallback value and reports each key of the flag config as a changed flag.

We believe this is the correct sdk behavior. The behavior allows the client app to decide what to do with the newly obtained flag config values. If you want to ignore the new flag config value, then you can certainly skip it in your implementation of `featureFlagDidUpdate`.

It may be that I do not fully understand the situation, and if I’ve missed something you were trying to communicate, then please reply with more information.

arun251:
Hi @natechiger, we're going to close this PR due to inactivity, but we can reopen if we get a reply.

PR #140, opened 2018-08-01
Add timer on mainRunLoop instead of currentRunLoop

brian-intuit:
I'm using polling and noticed that calling LDClient start on the main thread would block on the network call.  I want to do something like

DispatchQueue.global(qos: DispatchQoS.QoSClass.default).async {
    LDClient.sharedInstance().start(config, with: userBuilder)
}

When I did that, I noticed that the timer wasn't firing - I guess the thread for the currentRunLoop was already gone.

This change seems to solve my problem.

### DIFF
--- a/Darkly/LDThrottler.m
+++ b/Darkly/LDThrottler.m
@@ -76,7 +76,7 @@ const NSTimeInterval minDelayInterval = 1.0;
     NSDate *fireDate = [self.timerStartDate dateByAddingTimeInterval:delaySeconds];
 
     NSTimer *delayTimer = [[NSTimer alloc] initWithFireDate:fireDate interval:0 target:self selector:@selector(timerFired) userInfo:nil repeats:NO];
-    [[NSRunLoop currentRunLoop] addTimer:delayTimer forMode:NSDefaultRunLoopMode];
+    [[NSRunLoop mainRunLoop] addTimer:delayTimer forMode:NSDefaultRunLoopMode];
     return delayTimer;
 }
 


markpokornycos:
[Address PR 140](https://trello.com/c/4TjoP9wd/174-address-pr-140)

markpokornycos:
Thank you for your PR and recommendation. We’ve looked at your PR and have the following observations. 
•The code snippet that uses DispatchQueue will add the LDClient.start call to a global concurrent queue rather than the main thread.
•We tested the SDK version `2.13.2` to see if we could duplicate the situation you encountered. We were unable to duplicate, either running on the main thread or running on a global concurrent dispatch queue. Can you provide some sample code that demonstrates the issue?
•The Throttler timer is used to prevent multiple rapid calls to selected client components. In this case, setting the client online is throttled so that the LDClient doesn’t make multiple rapid requests for feature flags
•In polling mode, the LDClient requests feature flags right away when the LDClient is set online. A timer (different from the Throttler timer) is used to trigger the next feature flag request. The minimum interval is set to 5m, so even if the LDConfig has a smaller `pollingInterval`, the SDK will wait at least 5m before making the next flag request.
•Changing the throttler timer to the main run loop will cause any throttled runBlock to execute on the main thread. However any un-throttled runBlock would still execute on the thread the LDClient is running on. 
•Since you observed that changing the Throttler timer to fire on the main run loop affected the feature flag request, it seems that the LDClient is being set online rapidly more than 1 time. This is a condition that should be investigated and changed to prevent it from happening.
•We have started to examine whether to change the LDClient as you have suggested. Our initial impression is that the change would improve the LDClient by submitting throttled run blocks to the main thread if the LDClient was set online from a different thread. Since the intent is for the LDClient to run on the main thread, your recommendation seems appropriate. There may be other steps needed to fully enforce the LDClient running on the main thread.

brian-intuit:
You are on to something with your penultimate bullet.

What actually happened was that when I dispatched the start() call to a global queue, our app itself seemed to work fine.  But we have a bunch of integration tests that test our integration with LD and our feature flags themselves, and they all started failing.  Each time a test runs, it calls stop(), creates a new configuration, then calls start().

With the small change in this PR, I was happy to see lots of green check marks instead of red Xs.

arun251:
Hi @brian-intuit, do you still need this change, or were you able to resolve the issue by changing your tests?

brian-intuit:
I haven't found another way to resolve it, so would still like this fix to go in.

PR #165, opened 2019-02-26
Version 2.14.4

markpokornycos:


.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-# config file version: 1.0.3
+# config file version: 1.1.0
```
torchhound: Why did this jump a minor version?

Darkly.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>IDEWorkspaceSharedSettings_AutocreateContextsIfNeeded</key>
-	<true/>
+	<false/>
```
torchhound: Was this the fix for the schemes issue?

.circleci/config.yml
markpokornycos: Because we added steps to build all targets

Darkly.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
markpokornycos: in part. this prevents CocoaPods from creating schemes, which then cannot become public

PR #177, opened 2019-06-03
Version 3.0.4

JUSTINMKAUFMAN:
Added ability to pass an array of keys for feature flags that an application observes (see `trackedKeys` parameter in the `start:` method on `LDClient`).

By providing this array of keys, the SDK avoids performing maintenance on feature flags that the application does not track.


bwoskow-ld:
Hi Justin,

Thanks for taking the initiative to make these changes!

I'd like to know more about the problem you're seeing and trying to address. From your initial comment I understand that your iOS application is handling updates for flags that don't matter to your app -- in other words, your LD project/environment are configured such that only a subset of the flags accessible to your iOS app are actually in use by your iOS app. However, what is the problematic symptom you're seeing and trying to avoid? Battery usage, unnecessary storage, etc.?

JUSTINMKAUFMAN:
> Hi Justin,
> 
> Thanks for taking the initiative to make these changes!
> 
> I'd like to know more about the problem you're seeing and trying to address. From your initial comment I understand that your iOS application is handling updates for flags that don't matter to your app -- in other words, your LD project/environment are configured such that only a subset of the flags accessible to your iOS app are actually in use by your iOS app. However, what is the problematic symptom you're seeing and trying to avoid? Battery usage, unnecessary storage, etc.?

Thanks for the quick reply! I’ve been corresponding with your support team about a pretty constant stream of crashes our app has been experiencing over the past few weeks. Version 3.0.3 was intended to resolve these crashes, but unfortunately it did not. 

The motivation for this fork was to implement a stopgap measure for these crashes until a root cause is found and repaired (and in testing, it does appear to prevent the crashes we’ve been seeing).

I’d be happy to provide more detail on these crashes if you’re able to message (Slack?) or email me (justinkaufman@98point6.com).

bwoskow-ld:
Got it. I'm familiar with the problem you were seeing (leading to the 3.0.3 release) and am looped into your active support request. 

When you're ready to do so, I recommend sharing additional crash details on the support issue so that our support team can access them as well. I'll be able to see updates posted there.

JUSTINMKAUFMAN:
> Got it. I'm familiar with the problem you were seeing (leading to the 3.0.3 release) and am looped into your active support request.
> 
> When you're ready to do so, I recommend sharing additional crash details on the support issue so that our support team can access them as well. I'll be able to see updates posted there.

Just sent our crash logs in the support email thread.

JUSTINMKAUFMAN:
@bwoskow-ld I believe you guys have a memory leak in your fork of `neilco/EventSource`. Opening a PR in your fork with more info.

JUSTINMKAUFMAN:
See https://github.com/launchdarkly/ios-eventsource/pull/26

gwhelanLD:
Closing this due to inactivity. Feel free to reopen if this is still an issue.

Thanks,
@gwhelanLD

PR #180, opened 2019-06-19
Prepare 4.1.0 release

torchhound:
### Change
- Installs new `deviceModel` into `EnvironmentReporter` and renames old `deviceModel` to `deviceType`
- use `CwSysCtl` for macos model

### Fixed
- Fixed a concurrency bug that caused crashes in FlagStore.swift. This bug could surface during rapid updates to local flags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to the LaunchDarkly iOS SDK will be documented in this file.
 ### Multiple Environment clients
 Version 4.0.0 does not support multiple environments. If you use version `2.14.0` or later and set `LDConfig`'s `secondaryMobileKeys` you will not be able to migrate to version `4.0.0`. Multiple Environments will be added in a future release to the Swift SDK.
 
-## [4.0.1] - 2019-06-18
+## [4.1.0] - 2019-06-18
+### Change
+- Installs new `deviceModel` into `EnvironmentReporter` and renames old `deviceModel` to `deviceType`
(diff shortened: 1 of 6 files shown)


LaunchDarkly.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |ld|
 
   ld.name         = "LaunchDarkly"
-  ld.version      = "4.0.0"
+  ld.version      = "4.1.0"
```
gwhelanLD: What makes this a minor release rather than a patch? It looks like the `deviceModel`/`deviceType` changes are internal, but I might be wrong.

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to the LaunchDarkly iOS SDK will be documented in this file.
 ### Multiple Environment clients
 Version 4.0.0 does not support multiple environments. If you use version `2.14.0` or later and set `LDConfig`'s `secondaryMobileKeys` you will not be able to migrate to version `4.0.0`. Multiple Environments will be added in a future release to the Swift SDK.
 
+## [4.1.0] - 2019-06-19
+### Change
+- Installs new `deviceModel` into `EnvironmentReporter` and renames old `deviceModel` to `deviceType`
+- use `CwSysCtl` for macos model
```
gwhelanLD: Is this meant to be two separate entries? They seem related, but the `CrSysCtl` one is rather unclear on what it means.

PR #183, opened 2019-07-10
Only import WatchKit if running on watchOS

ugiacoman:
### Problem

In Xcode 11, you cannot import WatchKit on iOS simulators.

Error:
```
"WatchKit" is not available when building for iOS Simulator. Consider using `#if !os(iOS)` to conditionally import this framework.
```

### Solution

- Conditionally import WatchKit if running on watchOS

### Testing

- Passes Tests
- Runs on both Xcode 10.2 and Xcode 11

### DIFF
--- a/LaunchDarkly/LaunchDarkly/Service Objects/EnvironmentReporter.swift
+++ b/LaunchDarkly/LaunchDarkly/Service Objects/EnvironmentReporter.swift
@@ -8,8 +8,9 @@
 
 import Foundation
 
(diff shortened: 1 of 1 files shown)


torchhound:
Hi @ugiacoman, thanks for putting in a PR! I'll do some testing on this and get back to you.

torchhound:
Tested and LGTM!

torchhound:
This is not available in the iOS Client 4.1.2 release, thanks again for the PR!

Issue #186, opened 2019-07-10
Stack overflow when comparing two nil objects of type [String: Any]? 


DrBreen:
**Describe the bug**
Extension that overrides `==` operator for `Optional<[String:Any]>` causes infinite loop with stack overflow when comparing two `nil` optionals of this type.

**To reproduce**
1. Add LaunchDarkly 4.0 to Podfile:
`pod 'LaunchDarkly', '~>4.0.0'`
2. Run `pod install`
3. Add following code anywhere in application
```
let dict1: [String: Any]? = nil
let dict2: [String: Any]? = nil

if dict1 == dict2 {
    print("This will never be printed")
}
```
4. Run the application and make sure the code executes.
5. Application will crash due to stack overflow because of line 79 in `LaunchDarkly/Extensions/Dictionary.swift` - it enters recursive loop in case of `lhs` and `rhs` both being nil.

**Expected behavior**
`==` should return true for comparison of two `nil` `[String:Any]?`objects.

**SDK version**
4.0.0

**Language version, developer tools**
Swift 5

**OS/platform**
iOS 12

torchhound:
I just released iOS client 4.1.2 with your fix for this issue, thanks again for the PR!

lietusme:
This seems to be very similarly broken on 4.2.0, but now comparing [String: Any]? with nil:

`class Crash {
    static func test(parameters: [String: Any]? = nil) {
        if parameters == nil {
            print("Yes it's nil")
        }
    }
}`

Usage:
`Crash.test()`

Comparison overflows within this LD extension:
`extension Optional where Wrapped == [String: Any] {
    public static func == (lhs: [String: Any]?, rhs: [String: Any]?) -> Bool {
        guard let lhs = lhs
        else {
            **return rhs == nil**
        }
        guard let rhs = rhs
        else {
            return false
        }
        return lhs.isEqual(to: rhs)
    }

    public static func != (lhs: [String: Any]?, rhs: [String: Any]?) -> Bool {
        return !(lhs == rhs)
    }
}`

PR #187, opened 2019-07-10
Fix stack overflow when comparing two nil [String:Any]? objects

DrBreen:
* Bypass recursive call to `==` operator of `Optional<[String:Any]>` by using `guard` operator. See issue #186.


torchhound:
Hi @DrBreen, thanks for submitting this PR! I'll do some testing and get back to you here.

torchhound:
Tested and LGTM!

DrBreen:
Thanks for the quick response!

torchhound:
No problem, this fix will go out in the next release very soon!

PR #188, opened 2019-07-10
Fix build on Xcode 11

marcelofabri:
I was getting this error when building on Xcode 11:

```
ios-client-sdk/LaunchDarkly/LaunchDarkly/Service Objects/EnvironmentReporter.swift:13:8: error: no such module 'WatchKit'
import WatchKit
       ^
```


marcelofabri:
Ooops, seems that this is the same as  #183

torchhound:
This fix went out in iOS client 4.1.2, thanks for the PR!

Issue #220, opened 2020-08-31
Build warnings in SDK 5.1.0 with Xcode 12 beta 6


idrougge:
**Describe the bug**
`DiagnosticEvent.swift` has four warnings saying:
```
Immutable property will not be decoded because it is declared with an initial value which cannot be overwritten
```

**SDK version**
5.1.0

**Language version, developer tools**
Xcode 12, beta 6

torchhound:
Hi @idrougge, thanks for reporting this bug.

_Tracked internally as 88282._

gwhelanLD:
Fixed in the [5.2.0 release](https://github.com/launchdarkly/ios-client-sdk/releases/tag/5.2.0). Thanks again for reporting!

PR #227, opened 2020-10-08
fix: #220

ericlewis:
**Requirements**

- [X] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)
- [X] I have validated my changes against all supported platform versions

**Related issues**

#220 

Provide a clear and concise description of what you expect to happen. This PR fixes the warnings mentioned in #220, which were added to the Swift compiler [somewhat recently](https://github.com/apple/swift/pull/30218). The cause of the warning is setting default values for a Decodable objects properties. This causes subtle bugs where a given property may not be correctly decoded.

The solution proposed here is one of a few, simply settings the variable to a `var` while making its setter private. This retains the original behavior for those consuming the code, while also allowing Decodable to work properly.

Another option would be to inject the default values when `init` is called, but that adds more arguments and some models do not have initializers to begin with, this is a smaller change. Such a change would look like this:
```swift
struct DiagnosticInit: DiagnosticEvent, Codable {
    let kind: DiagnosticKind
    let id: DiagnosticId
    let creationDate: Int64

    let sdk: DiagnosticSdk
    let configuration: DiagnosticConfig
    let platform: DiagnosticPlatform

    init(config: LDConfig, diagnosticKind: DiagnosticKind = .diagnosticInit, diagnosticId: DiagnosticId, creationDate: Int64) {
        self.kind = diagnosticKind
        self.id = diagnosticId
        self.creationDate = creationDate

        self.sdk = DiagnosticSdk(config: config)
        self.configuration = DiagnosticConfig(config: config)
        self.platform = DiagnosticPlatform(config: config)
    }
}
```

This seems possibly like something we don't really want, because it opens up the user to being able to change a value that seems like it should remain static.

**Additional context**

This is the background for the warning: https://github.com/apple/swift/pull/30218

### DIFF
--- a/LaunchDarkly/LaunchDarkly/Models/DiagnosticEvent.swift
+++ b/LaunchDarkly/LaunchDarkly/Models/DiagnosticEvent.swift
@@ -19,7 +19,7 @@ protocol DiagnosticEvent {
 }
 
 struct DiagnosticInit: DiagnosticEvent, Codable {
-    let kind = DiagnosticKind.diagnosticInit
+    private(set) var kind = DiagnosticKind.diagnosticInit
     let id: DiagnosticId
     let creationDate: Int64
 
@@ -38,7 +38,7 @@ struct DiagnosticInit: DiagnosticEvent, Codable {
 }
 
 struct DiagnosticStats: DiagnosticEvent, Codable {
-    let kind = DiagnosticKind.diagnosticStats
+    private(set) var kind = DiagnosticKind.diagnosticStats
     let id: DiagnosticId
     let creationDate: Int64
 
@@ -66,7 +66,7 @@ struct DiagnosticId: Codable {
 }
 
 struct DiagnosticPlatform: Codable {
-    let name: String = "swift"
+    private(set) var name: String = "swift"
     let systemName: String
     let systemVersion: String
     let backgroundEnabled: Bool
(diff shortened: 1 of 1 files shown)


gwhelanLD:
Hi @ericlewis,

We appreciate your submission of this PR!  We actually have some internal changes to fix these warnings on the way to release currently. The structs with default fields will never actually need to be decoded in SDK use, so the approach we've done internally is to remove the derived `Decodable` protocol conformance leaving the structs only `Encodable`. This also requires some test changes as there are some tests for encoding/decoding of these structs, which are especially overboard given we never decode any of the ones with constant properties. The `DiagnosticEvent` models were the first significant place to use the `Codable` protocols in the SDK, so the tests were mainly to improve my understanding. 

I agree that your approach here is cleaner than providing construction arguments for the fields, but I'm leaning towards the internal changes to keep these fields as `let`s and remove the redundant testing code. Happy to hear any feedback on that approach as well though, we on the SDK team have to wear a lot of hats so we're not always domain experts on the specific platforms and styles. 

Thanks for the very thoughtful PR, we really value external contributions to our SDKs. We should be able to roll out a release with these warnings fixed soon.

Thanks again,
@gwhelanLD

ericlewis:
@gwhelanLD I was exactly curious if it *needed* to be Decodable 💯 
I would think that is certainly the better solution. Closing this for now.

PR #246, opened 2021-04-29
Fixing crashes when using overflowing values

delannoyk:
**Requirements**

- [x] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)

**Related issues**

I didn't create a GitHub issue for this, but happy to if necessary.

`Int(exactly:)` will return nil in case of an overflow whereas `Int(_:)` will just crash. When converting the `Double` to an `Int`, if it fails, we now use `Int.max` as a rescue.

Fixing the crash was I think necessary. That being said, we only noticed that crash because we wanted to disable updates to our feature flags during a user session (we want to have a consistent experience throughout the session). There's no way that I could find to do that right now, so we opted for having a _really_ long update interval.
Adding this feature would have been nice too, but was more work that I intended.

**Additional context**

Nothing else from me.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/Models/DiagnosticEvent.swift
+++ b/LaunchDarkly/LaunchDarkly/Models/DiagnosticEvent.swift
@@ -124,12 +124,12 @@ struct DiagnosticConfig: Codable {
         customEventsURI = config.eventsUrl != LDConfig.Defaults.eventsUrl
         customStreamURI = config.streamUrl != LDConfig.Defaults.streamUrl
         eventsCapacity = config.eventCapacity
-        connectTimeoutMillis = Int(round(config.connectionTimeout * 1_000))
-        eventsFlushIntervalMillis = Int(round(config.eventFlushInterval * 1_000))
+        connectTimeoutMillis = Int(exactly: round(config.connectionTimeout * 1_000)) ?? .max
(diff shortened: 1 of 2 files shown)


torchhound:
Hi @delannoyk thanks for submitting this PR! We will review this internally to check if this fits with intended SDK behavior.

bwoskow-ld:
Hi @delannoyk,

While our SDK team hasn't yet had a chance to look at this issue, I did want to recommend that you use smaller intervals if you haven't already done so. Our SDK isn't really designed to operate with "_really_ long" intervals because, depending on which intervals you extends, flag values will be stale (which is actually your intent here) and the event buffer is more likely to overflow and start dropping events.

If you choose to proceed with large intervals, consider setting them to the max int value so as to avoid running into this crash.

Cheers,
@bwoskow-ld

gwhelanLD (approved):
Looks great, thanks! We'll get a patch out with this change soon.

Thanks,
@gwhelanLD

gwhelanLD:
Thanks again for the contribution! It's been included in the [5.4.2 release](https://github.com/launchdarkly/ios-client-sdk/releases/tag/5.4.2).

Thanks,
@gwhelanLD

delannoyk:
Thank you for the update 👍🏻

Issue #257, opened 2021-10-12
LDSwiftSourceEvent link issue


kinglezhuang:
I open workspace and build ios framework. Then embedded the LanuchDarkly.framework & LDSwiftEventSource.framework into our project. It show below error:
Class _TtC18LDSwiftEventSource11EventParser is implemented in both /Users/rcadmin/jenkins-slave/workspace/CommonCI-mThor-Core-Auto-Generate/RingCentral-Core/build/Build/Products/Debug-iphonesimulator/core-test.xctest/Frameworks/LDSwiftEventSource.framework/LDSwiftEventSource (0x170ba2498) and /Users/rcadmin/jenkins-slave/workspace/CommonCI-mThor-Core-Auto-Generate/RingCentral-Core/build/Build/Products/Debug-iphonesimulator/core-test.xctest/Frameworks/LaunchDarkly.framework/LaunchDarkly (0x1708fa870). One of the two will be used. Which one is undefined.

I found EventParser symbol exist in LanuchDarkly.framework, not reasonable because it just belongs to LDSwiftEventSource.framework.

I open project & found the dependency is framework, but link by using static in project build  phase tab.
Incorrect, right?

louis-launchdarkly:
Hello @kinglezhuang, Apologies for the delay and thank you for waiting. From your error, it seems that the linker is complaining there are multiple implementations of LDSwiftEventSource. Because the iOS SDK already listed LDSwiftEventSource as a [dependency] (https://github.com/launchdarkly/ios-client-sdk/blob/v5/Package.swift#L22), have you tried to only embed `LaunchDarkly.framework`?

https://github.com/launchdarkly/hello-ios is a helpful reference on how to use the iOS SDK.

If the suggestions above do not resolve your issue, please contact our support via https://support.launchdarkly.com/hc/en-us

louis-launchdarkly:
No more comments from the requester.

PR #271, opened 2022-06-14
Avoid exposing LDSwiftEventSource as dependency, when building xcframework

dingtianran:
**Requirements**

- [/] I have added test coverage for new or changed functionality
- [/] I have followed the repository's [pull request submission guidelines](../blob/v6/CONTRIBUTING.md#submitting-pull-requests)
- [/] I have validated my changes against all supported platform versions

**Related issues**

Due to some users can't using public SPM on GitHub, using pre-build binary xcframework is a good option. But at the source code, using `import LDSwiftEventSource` will expose it as another dependency(both `LaunchDarkly.xcframework` and `LDSwiftEventSource.xcframework`) which isn't necessary.

Potential related issue: https://github.com/launchdarkly/ios-client-sdk/issues/257

Solution👉 Mark `LDSwiftEventSource` as @ _implementationOnly will prevent it from expose as dependency: it'll remove LDSwiftEventSource from the final .swiftinterface file

Screen Shot 2022-06-14 at 20 52 13

### DIFF
--- a/LaunchDarkly/GeneratedCode/mocks.generated.swift
+++ b/LaunchDarkly/GeneratedCode/mocks.generated.swift
@@ -2,7 +2,7 @@
 // DO NOT EDIT
 
 import Foundation
-import LDSwiftEventSource
+@_implementationOnly import LDSwiftEventSource
 @testable import LaunchDarkly
 
 // swiftlint:disable large_tuple
--- a/LaunchDarkly/LaunchDarkly/Networking/DarklyService.swift
+++ b/LaunchDarkly/LaunchDarkly/Networking/DarklyService.swift
@@ -1,5 +1,5 @@
 import Foundation
-import LDSwiftEventSource
+@_implementationOnly import LDSwiftEventSource
 
 typealias ServiceResponse = (data: Data?, urlResponse: URLResponse?, error: Error?)
 typealias ServiceCompletionHandler = (ServiceResponse) -> Void
--- a/LaunchDarkly/LaunchDarkly/ServiceObjects/ClientServiceFactory.swift
+++ b/LaunchDarkly/LaunchDarkly/ServiceObjects/ClientServiceFactory.swift
@@ -1,5 +1,5 @@
 import Foundation
-import LDSwiftEventSource
+@_implementationOnly import LDSwiftEventSource
 
 protocol ClientServiceCreating {
     func makeKeyedValueCache(cacheKey: String?) -> KeyedValueCaching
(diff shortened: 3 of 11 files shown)


keelerm84:
Thank you for your contribution, and my apologies for the delay in
responding.

According to the [Swift docs][1], `@_implementationOnly` isn't a stable language feature and it's use is _highly_ discouraged.

I would like to gain more insight into your setup though as this background helps inform our future changes.

Can you explain the nature of your setup a bit more? What type of application are you building, what does your build process look like, which platforms are you targeting?
   
My hope is to replicate your setup locally so that if this feature becomes stable, I can add regression testing for it.
   
[1]: https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md

dingtianran:
Hi @keelerm84 thank you for your reply, here is my use case(a bit odd I have to admit). Due to security requirements, I can't use public github/swift package link in my Xcode project. The way I managed to do is manually have a bunch of xcframeworks referred in project, instead of SPM denpendencies. If the library I need can provide pre-built xcframeworks(like [Firebase](https://github.com/firebase/firebase-ios-sdk/releases/latest)), that's sweet for me. If not, then I compile them from source code to binary xcframeworks then chuck into Xcode project(like LaunchDarkly).

During the compilation, I found that I have to build both `LaunchDarkly.xcframework` and `LDSwiftEventSource.xcframework`, and put both of them into project, because LaunchDarkly depend on LDSwiftEventSource. But in my opinion this is unnecessary, as a consumer of LaunchDarkly, I want `LaunchDarkly.xcframework` to be the single source of truth, I don't want have another sidekick library. This is why I did some research and found `@_implementationOnly` could help.

What you concerned of is legit, this `@_implementationOnly` is not recommended yet by swift community to be using publicly. Maybe there is another way to produce an independent `LaunchDarkly.xcframework`?

dlewandaDK:
Bumping the conversation on this, as I think something like this will help my team as well. We are trying to use Xcode Cloud, but because `LDSwiftEventSource` is a dynamic library, Xcode Cloud runs into [this issue](https://developer.apple.com/forums/thread/689907). If there's a way to hide `LDSwiftEventSource` such that it doesn't run into this issue, it will help at least until Apple can rectify this problem with their CI solution.

rayalarajee:
Hi , Facing same issue while creating the xcframework,  any solution will help . Thanks in advance.

dingtianran:
@rayalarajee the solution is add @_implementationOnly before all "import LDSwiftEventSource" and build xcframework. This will make a standalone xcframework, no other dependency needed. It's a hack but works well for me

rayalarajee:
Thank you. @dingtianran

PR #308, opened 2023-09-12
Update Package.swift to support swift-tools-version:5.8

pierre-wehbe:
**Requirements**

- [X] I have followed the repository's [pull request submission guidelines](../blob/v6/CONTRIBUTING.md#submitting-pull-requests)
- [X] I have validated my changes against all supported platform versions

**Related issues**

swift-tools-version:5.8 have deprecated the use of .package(name:xxx, url:xxx, from:xxx) to .package(url:xxx, from:xxx).
This causes our internal SPM using 5.8 to break

Update LD SPM to support 5.8

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.8
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


pierre-wehbe:
In case 5.8 is a concern (since it becomes the minimum), I believe keeping 5.2 would still work

tanderson-ld:
Hi @pierre-wehbe , thanks for opening this PR.  One of our team will take a look to verify this has no adverse effects on our packages.  Thank you!

tanderson-ld:
HI @pierre-wehbe .  We will be releasing support for XCode15 in the near future.  During this work, we looked at the swift-tools-version you have proposed.  Unfortunately that bump would be too drastic for our customer base at the moment.  We are going to be bumping it from 5.2 to 5.3 however.  Are you able to have your build process ignore the deprecation?

PR #322, opened 2023-12-28
ci: Automate release process with release please

keelerm84:


.github/actions/ci/action.yml
```diff
@@ -0,0 +1,83 @@
...
+      shell: bash
+      run: make build-contract-tests
+
+    - name: Start contract tests in background
+      shell: bash
+      run: make start-contract-test-service-bg
+
+    - name: Run contract tests
+      shell: bash
+      run: sleep 5 && make run-contract-tests
```
cwaldren-ld, line 85: Explanation of the sleep would be good

.github/actions/publish/action.yml
```diff
@@ -0,0 +1,14 @@
...
+    description: 'Is this a dry run. If so no package will be published.'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Push to cocoapods
+      if: ${{ inputs.dry_run == 'false' }}
+      shell: bash
+      run: pod trunk push LaunchDarkly.podspec --allow-warnings --verbose
```
cwaldren-ld, line 14: Guessing the tool itself doesn't have a dry run mode?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
+name: Run CI
+on:
+  push:
+    branches: [ v8 ]
+    paths-ignore:
+      - '**.md' # Do not need to run CI for markdown changes.
+  pull_request:
+    branches: [ v8 ]
+    paths-ignore:
+      - '**.md'
```
cwaldren-ld, line 10: Wondering this for other SDKs as well. Do you think it should run daily/weekly so that incompatibilities don't slip in (like the Github runner tools subtly changing?)

Makefile
```diff
@@ -1,19 +1,19 @@
...
+	cd ./ContractTests && swift run ContractTests serve -b 0.0.0.0:8080
 
 start-contract-test-service-bg:
 	@echo "Test service output will be captured in $(TEMP_TEST_OUTPUT)"
 	@make start-contract-test-service >$(TEMP_TEST_OUTPUT) 2>&1 &
 
 run-contract-tests:
 	@curl -s https://raw.githubusercontent.com/launchdarkly/sdk-test-harness/master/downloader/run.sh \
-      | VERSION=v2 PARAMS="-url http://localhost:8080 -debug -stop-service-at-end -skip-from ./ContractTests/testharness-suppressions.txt $(TEST_HARNESS_PARAMS)" sh
+      | VERSION=v2 PARAMS="-url http://0.0.0.0:8080 -debug -stop-service-at-end -skip-from ./ContractTests/testharness-suppressions.txt $(TEST_HARNESS_PARAMS)" sh
```
cwaldren-ld: Was localhost not working?

cwaldren-ld (approved):

tanderson-ld (approved):

.github/workflows/ci.yml
keelerm84: Yes, I was actually thinking about this as well. This is my last PR for release please in my repos, so I'm going to work on getting this one merged, and then probably come back through everyone one final time.

Makefile
keelerm84: It might actually work. I was running into some other problems originally with this and tried that first. I'll revert and see if it runs through.

.github/actions/publish/action.yml
keelerm84: Correct. I did check originally but it doesn't appear to support it. We do perform a pod spec lint every time and that pretty much does everything except push so we should be _somewhat_ shielded.

.github/actions/ci/action.yml
keelerm84: Added.

tanderson-ld (approved):

PR #334, opened 2024-02-08
fix: Add privacy manifest 

keelerm84:


tanderson-ld (approved):

choulepoka:
Is there a way to stop the _tracking_ of users? Because tracking is a big deal in the Apple world, if the LD sdk tracks, when we have to request user's permission, and if the user says no, we gotta _stop_ tracking.

Is there a way to do that without crippling the LD integration? @keelerm84

keelerm84:
> Is there a way to stop the _tracking_ of users? Because tracking is a big deal in the Apple world, if the LD sdk tracks, when we have to request user's permission, and if the user says no, we gotta _stop_ tracking.
> 
> Is there a way to do that without crippling the LD integration? @keelerm84

My apologies for the delay in getting back to your question!

It is my understanding that the type of data our SDK uses does not require user interaction for approval. We are not tracking users for advertisement purposes; rather, any identifier is consumer provided and sent to our servers for the purpose of flag evaluation and experimentation.

choulepoka:
@keelerm84 If that is truly the case, then the following values should be `false`:
- `NSPrivacyTracking` 
- `NSPrivacyCollectedDataTypeTracking`

Otherwise, you are basically declaring to Apple that the LaunchDarkly Framework is tracking for the purpose of App Tracking Transparency.

keelerm84:
@choulepoka Yes, you are correct. We had the type tracking field set incorrectly. This is being adjusted in https://github.com/launchdarkly/ios-client-sdk/pull/366 and should be out in the next patch release (9.6.2).

PR #335, opened 2024-02-08
build: Favor pod lib lint over pod spec lint

keelerm84:


keelerm84:
`pod spec` requires the tag in the spec file to actually exist upstream. That doesn't help when you are making a change that only exists locally. So I'm switching to this.

tanderson-ld (approved):

PR #341, opened 2024-02-12
feat: Add new identify method with time out support

keelerm84:
A callback provided to the `identify` method may not be executed for a
long time under the right circumstances (e.g. the network is slow or
unavailable).

Customers can now ensure their callbacks will fire within a specified
time interval by calling the new
`LDClient.identify(context:timeOut:completion:)` method.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -286,6 +286,38 @@ public class LDClient {
         }
     }
 
+    /**
+     The LDContext set into the LDClient may affect the set of feature flags returned by the LaunchDarkly server, and ties event tracking to the context. See `LDContext` for details about what information can be retained.
+
+     Normally, the client app should create and set the LDContext and pass that into `start(config: context: completion:)`.
+
+     The client app can change the active `context` by calling identify with a new or updated LDContext. Client apps should follow [Apple's Privacy Policy](apple.com/legal/privacy) when collecting user information.
+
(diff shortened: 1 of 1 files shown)


louis-launchdarkly (approved):

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -286,6 +286,38 @@ public class LDClient {
...
+            completion(true)
+        }
+
+        identify(context: context) {
+            guard !cancel else { return }
+
+            cancel = true
+            completion(false)
+        }
+    }
```
tanderson-ld, line 319: Is it acceptable that the timeout completion will be called with true, but the identify may still be internally processing and may even succeed with no successful completion callback?  I believe this case is possible.

tanderson-ld, line 319: Can you verify the following case, I'm unsure if this is an issue.

identify(context: myContextA, timeout: 10 seconds, completion: completionA)
identify(context: myContextB, timeout: 9.9 seconds, completion: completionB)

* assume identify of myContextA takes more than 0.1 seconds.

I believe completionA will be called with success, and completionB will be called with a timeout, but myContextB will be the active context.  I think the `internalIdentifyQueue` is involved here.

tanderson-ld:
I think using @kinyoklion's 1 identify in progress spec may be the only course of action to add the timeout support on top of the existing code (which does not support cancellation of identify in progress).

keelerm84:
Going to close this. Originally was thinking we might be able to make some concessions on the first round of this, but have decided against it.

PR #343, opened 2024-02-20
feat: Implement shedding identity queue

keelerm84:
Previously, customers could queue a boundless limit of identify requests. The SDK would work its way through this FIFO queue, processing all intermediate but unnecessary requests.

With this change, intermediate identify requests will be shed from the processing queue.

NOTE: To preserve backwards compatibility, the original identify method will queue up "unsheddable" tasks which will continue to queue as before. Usage of the new `identify` method will allow developers to opt-in to this new behavior.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -278,15 +278,53 @@ public class LDClient {
      - parameter context: The LDContext set with the desired context.
      - parameter completion: Closure called when the embedded `setOnlineIdentify` call completes, subject to throttling delays. (Optional)
     */
+    @available(*, deprecated, message: "Use LDClient.identify(context: completion:) with non-optional completion parameter")
     public func identify(context: LDContext, completion: (() -> Void)? = nil) {
-        let dispatch = DispatchGroup()
-        LDClient.instances?.forEach { _, instance in
-            dispatch.enter()
-            instance.internalIdentify(newContext: context, completion: dispatch.leave)
+        _identify(context: context, sheddable: false) { _ in
+            if let completion = completion {
+                completion()
+            }
         }
-        if let completion = completion {
-            dispatch.notify(queue: DispatchQueue.global(), execute: completion)
+    }
+
+    /**
+     The LDContext set into the LDClient may affect the set of feature flags returned by the LaunchDarkly server, and ties event tracking to the context. See `LDContext` for details about what information can be retained.
+
+     Normally, the client app should create and set the LDContext and pass that into `start(config: context: completion:)`.
+
+     The client app can change the active `context` by calling identify with a new or updated LDContext. Client apps should follow [Apple's Privacy Policy](apple.com/legal/privacy) when collecting user information.
+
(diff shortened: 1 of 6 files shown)


LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -278,15 +278,53 @@ public class LDClient {
      - parameter context: The LDContext set with the desired context.
      - parameter completion: Closure called when the embedded `setOnlineIdentify` call completes, subject to throttling delays. (Optional)
     */
+    @available(*, deprecated, message: "Use LDClient.identify(context: completion:) with non-optional completion parameter")
```
keelerm84, line 281: Deprecating this method so that customers will start moving over to the callback that informs them of success or failure.

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -278,15 +278,53 @@ public class LDClient {
...
+
+     When a new context is set, the LDClient goes offline and sets the new context. If the client was online when the new context was set, it goes online again, subject to a throttling delay if in force (see `setOnline(_: completion:)` for details). A completion may be passed to the identify method to allow a client app to know when fresh flag values for the new context are ready.
+
+     While only a single identify request can be active at a time, consumers of this SDK can call this method multiple times. To prevent unnecessary network traffic, these requests are placed
+     into a sheddable queue. Identify requests will be shed if 1) an existing identify request is in flight, and 2) a third identify has been requested which can be replace the one being shed.
+
+     - parameter context: The LDContext set with the desired context.
+     - parameter completion: Closure called when the embedded `setOnlineIdentify` call completes, subject to throttling delays. (Optional)
+     */
+    public func identify(context: LDContext, completion: @escaping (_ result: IdentifyResult) -> Void) {
```
keelerm84, line 305: The other identify method has an optional completion method. This one isn't optional.

If I continue with the completion method being optional, then the method isn't fully disambiguated. Customers could have code like: `LDClient.get()?.identify(context: context, completion: nil)` and the compiler wouldn't know which one we meant.

Given that these identify requests might never actually complete, it seems kind of reasonable to force customers listen for those results, just as a measure of good practice for how to use this SDK. What do you think?

If we decide to make it optional, how do we deal with the backwards compatibility naming convention? Keep in mind that if we name it something like `identifyWithShedding` then in the next major release, customers who took action based on the deprecation notice are going to have to change their code AGAIN.

We could consider making it non-nullable in this. version and loosen that restriction in the next. Or we could wait to see what customer feedback is and add the nullability back at that time.

kinyoklion: I am interested to see what customer feedback is. It doesn't seem unreasonable at this point to require it.

kinyoklion:
Can/should the contract tests be updated to use the shedding version?

kinyoklion (approved):

LaunchDarkly/LaunchDarkly/LDClient.swift
tanderson-ld: They can always pass a no-op and ignore it.

tanderson-ld (approved):

PR #344, opened 2024-02-20
feat: Add new identify method with time out support

keelerm84:
A callback provided to the `identify` method may not be executed for a long time under the right circumstances (e.g. the network is slow or unavailable).

Customers can now ensure their callbacks will fire within a specified time interval by calling the new
`LDClient.identify(context:timeout:completion:)` method.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -321,10 +321,40 @@ public class LDClient {
         }
 
         let identifyTask = Task(work: work, sheddable: sheddable) { [self] result in
-            os_log("%s identity completion with result %s", log: config.logger, type: .debug, typeName(and: #function), String(describing: result))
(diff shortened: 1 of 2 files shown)


LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -327,6 +327,36 @@ public class LDClient {
...
+
+            cancel = true
+            completion(.timeout)
+        }
+
+        identify(context: context) { result in
+            guard !cancel else { return }
+
+            cancel = true
+            completion(result)
```
tanderson-ld, line 356: Is it possible that both the timeout async call and this call get past the guard clause at the same time?

keelerm84: I do not believe so. Both the timeout and the identify completion code rely on the global DispatchQueue. So the requests should always be appropriately ordered.

tanderson-ld (approved):

PR #355, opened 2024-03-15
feat: Honor polling interval between restarts

keelerm84:
If an application receives a flag update and then is closed, only to be opened back up, we shouldn't need to make another polling request immediately as the data is within the acceptable freshness threshold.

The same is true when the app is put in the background and brought forward again.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
+++ b/LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
@@ -5,7 +5,7 @@ protocol FeatureFlagCaching {
     // sourcery: defaultMockValue = KeyedValueCachingMock()
     var keyedValueCache: KeyedValueCaching { get }
 
-    func getCachedData(cacheKey: String) -> (items: StoredItems?, etag: String?)
+    func getCachedData(cacheKey: String) -> (items: StoredItems?, etag: String?, lastUpdated: Date?)
 
     // When we update the cache, we save the flag data and if we have it, an
     // etag. For polling, we should always have the flag data and an etag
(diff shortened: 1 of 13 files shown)


LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
```diff
@@ -42,16 +42,24 @@ final class FeatureFlagCache: FeatureFlagCaching {
...
-        return (items: cachedFlags.flags, etag: etag)
+        var cachedContexts: [String: Int64] = [:]
+        if let cacheMetadata = keyedValueCache.data(forKey: "cached-contexts") {
+            cachedContexts = (try? JSONDecoder().decode([String: Int64].self, from: cacheMetadata)) ?? [:]
+        }
+
+        guard let lastUpdated = cachedContexts[cacheKey]
+        else { return (items: cachedFlags.flags, etag: etag, lastUpdated: nil) }
+
+        return (items: cachedFlags.flags, etag: etag, lastUpdated: Date(timeIntervalSince1970: TimeInterval(lastUpdated / 1_000)))
```
tanderson-ld, line 73: Just to confirm here, the lastUpdate from the cache is millis, but the TimeInterval takes in seconds as a float?

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
```diff
@@ -5,7 +5,7 @@ protocol FeatureFlagCaching {
     // sourcery: defaultMockValue = KeyedValueCachingMock()
     var keyedValueCache: KeyedValueCaching { get }
 
-    func getCachedData(cacheKey: String) -> (items: StoredItems?, etag: String?)
+    func getCachedData(cacheKey: String) -> (items: StoredItems?, etag: String?, lastUpdated: Date?)
```
tanderson-ld, line 19: Missing documentation.  The other method also is missing documentation around lastUpdated.

LaunchDarkly/LaunchDarkly/ServiceObjects/FlagSynchronizer.swift
```diff
@@ -151,9 +154,15 @@ class FlagSynchronizer: LDFlagSynchronizing, EventHandler {
...
         }
 
+        if let lastTime = self.lastCachedRequestedTime {
+            let fireAt = lastTime.addingTimeInterval(pollingInterval)
+            flagRequestTimer = LDTimer(withTimeInterval: pollingInterval, fireQueue: syncQueue, fireAt: fireAt, execute: processTimer)
+            syncQueue.async { [self] in reportSyncComplete(.upToDate) }
+        } else {
+            flagRequestTimer = LDTimer(withTimeInterval: pollingInterval, fireQueue: syncQueue, execute: processTimer)
+            makeFlagRequest(isOnline: true)
+        }
```
tanderson-ld, line 165: It seems like it should be possible to make these two code paths converge.  If the fireAt is in the past, does the LDTimer fire immediately?

If so, seems like defaulting last cached requested time to epoch when it is missing should work.

Perhaps I'm missing why makeFlagRequest(...) is in one path but not the other and that reason I'm missing is why the paths can't converge?

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
keelerm84: These methods are internal and so won't be reflected in the public documentation. I have still added some additional documentation.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
keelerm84: Yes. 

A [TimeInterval](https://developer.apple.com/documentation/foundation/timeinterval) is always in seconds.

You can [see here](https://github.com/launchdarkly/ios-client-sdk/blob/099ffe0dfab9a6355652529ce3ff89aa3afce756/LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift#L71) that the timestamp in the cache is millis.

LaunchDarkly/LaunchDarkly/ServiceObjects/FlagSynchronizer.swift
keelerm84: It was affecting unit test execution order. I've updated the unit tests and combined these two code paths.

tanderson-ld (approved):

PR #362, opened 2024-04-01
fix: Add missing privacy tracking fields in manifest

keelerm84:
The [documentation](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files) seems to suggest you should set all 4 of these fields. So I'm doing that for completeness.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/PrivacyInfo.xcprivacy
+++ b/LaunchDarkly/LaunchDarkly/PrivacyInfo.xcprivacy
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
 	<key>NSPrivacyCollectedDataTypes</key>
 	<array>
 		<dict>


tanderson-ld (approved):

louis-launchdarkly (approved):

LaunchDarkly/LaunchDarkly/PrivacyInfo.xcprivacy
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
```
kinyoklion, line 7: > If you set NSPrivacyTracking to true then you need to provide at least one internet domain in NSPrivacyTrackingDomains; otherwise, you can provide zero or more domains.

Is this meant to be for analytic events?

rsoberano-ld: I think we should be okay here, the definition of tracking from the site doesn't seem to apply to analytics events: 

> Tracking refers to the act of linking user or device data collected from your app with user or device data collected from other companies’ apps, websites, or offline properties for targeted advertising or advertising measurement purposes. Tracking also refers to sharing user or device data with data brokers.

rsoberano-ld (approved):

LaunchDarkly/LaunchDarkly/PrivacyInfo.xcprivacy
kinyoklion: I misread the other property.

PR #364, opened 2024-04-01
fix: Deprecate calling `start` without a timeout parameter

keelerm84:
In addition to this deprecation, calling `start` or `identify` with a timeout value in excess of 15 seconds will result in a warning message being logged.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -42,6 +42,9 @@ public class LDClient {
 
     private static var instances: [String: LDClient]?
 
+    // If the SDK is provided a timeout value that exceeds this value, a warning will be logged.
+    private static let longTimeoutInterval: TimeInterval = 15
+
     /**
      Reports the online/offline state of the LDClient.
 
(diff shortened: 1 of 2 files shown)


tanderson-ld (approved):

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -753,6 +761,10 @@ public class LDClient {
     - parameter completion: Closure called when the embedded `setOnline` call completes. Takes a Bool that indicates whether the completion timedout as a parameter. (Optional)
```
mmrj, line 761: Line 760 could be a good spot to include additional info per @louis-launchdarkly 's [suggestion](https://launchdarkly.slack.com/archives/CSGD4E4LB/p1712012043905799):

```
    - parameter startWaitSeconds: A TimeInterval that determines when the completion will return if no flags have been returned from the network. If you use a large TimeInterval and wait for the timeout, then any network delays will cause your application to wait a long time before continuing execution.
```

keelerm84: Done!

PR #366, opened 2024-04-03
fix: Mark  NSPrivacyCollectedDataTypeTracking as false in privacy manifest

keelerm84:
According to
https://developer.apple.com/app-store/user-privacy-and-data-use/

> Tracking refers to the act of linking user or device data collected
from your app with user or device data collected from other companies’ apps, websites, or offline properties for targeted advertising or advertising measurement purposes.

The data we collect is not used for advertising.


tanderson-ld (approved):
I missed this during previous reviews.  Lots of new terminology.

louis-launchdarkly (approved):

PR #369, opened 2024-04-10
chore: add .sdk_metadata.json

cwaldren-ld:


.sdk_metadata.json
```diff
@@ -0,0 +1,12 @@
...
+  "sdks": {
+    "swift-client-sdk": {
+      "name": "iOS SDK",
+      "type": "client-side",
+      "languages": [
+        "Swift"
+      ]
+    }
+  }
+}
```
cwaldren-ld, line 12: So, I've given this the primary key `swift-client-sdk` because I know we have been thinking about referring to it like that for a while. This isn't actually user facing, just the key used in the metadata database.

We can eventually update the `name` if we want, too.

Finally, I've put `Swift` as the language - should it have `Objective-C` too?

tanderson-ld: I think the objective c layer is generated.  https://github.com/launchdarkly/ios-client-sdk/blob/v9/LaunchDarkly/LaunchDarkly/ObjectiveC/ObjcLDClient.swift

tanderson-ld (approved):

.sdk_metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1,
+  "sdks": {
+    "swift-client-sdk": {
+      "name": "iOS SDK",
+      "type": "client-side",
+      "languages": [
+        "Swift"
```
cwaldren-ld:
```suggestion
        "Swift", "Objective-C"
```

PR #370, opened 2024-04-19
feat: Add initial support for hooks

keelerm84:


LaunchDarkly/LaunchDarkly/Models/Hooks/Hook.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+/// Implementation specific hook data for evaluation stages.
+///
+/// Hook implementations can use this to store data needed between stages.
+public typealias EvaluationSeriesData = [String: Any]
+
+/// Protocol for extending SDK functionality via hooks.
+public protocol Hook {
```
tanderson-ld, line 9: Is this forward compatible?  I'm not super familiar with swift protocols.

LaunchDarkly/LaunchDarklyTests/LDClientHookSpec.swift
```diff
@@ -0,0 +1,131 @@
...
+import XCTest
+@testable import LaunchDarkly
+
+final class LDClientHookSpec: XCTestCase {
+//    final class LDContextSpec: XCTestCase {
+//    override func spec() {
+//        registrationSpec()
+//        evaluationOrderSpec()
+//        dataPassingSpec()
+//    }
```
tanderson-ld: Remove commented out code if it can be.

LaunchDarkly/LaunchDarkly/LDClientVariation.swift
```diff
@@ -143,32 +143,59 @@ extension LDClient {
...
-            if featureFlag.value == .null {
-                result = LDEvaluationDetail(value: defaultValue, variationIndex: featureFlag.variation, reason: featureFlag.reason)
-            } else {
-                do {
-                    let convertedValue = try LDValueDecoder().decode(T.self, from: featureFlag.value)
-                    result = LDEvaluationDetail(value: convertedValue, variationIndex: featureFlag.variation, reason: featureFlag.reason)
-                } catch let error {
-                    os_log("%s type conversion error %s: failed converting %s to type %s", log: config.logger, type: .debug, typeName(and: #function), String(describing: error), String(describing: featureFlag.value), String(describing: T.self))
-                    result = LDEvaluationDetail(value: defaultValue, variationIndex: nil, reason: ["kind": "ERROR", "errorKind": "WRONG_TYPE"])
+        return evaluateWithHooks(flagKey: flagKey, defaultValue: defaultValue, methodName: "variationDetailInternal") {
```
tanderson-ld: I believe the method name should be the top level method that the customer invokes on the SDK.  Such as `LDClientVariation. stringVariation`.  I think `variationDetailInternal` needs to be changed?  Ryan may need to chime in here.

The spec has `method (string/enumeration, readonly, required): The method being executed, for instance LDClient.variationBool.`

LaunchDarkly/LaunchDarkly/Models/Hooks/Hook.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+/// Implementation specific hook data for evaluation stages.
+///
+/// Hook implementations can use this to store data needed between stages.
+public typealias EvaluationSeriesData = [String: Any]
```
tanderson-ld, line 6: This should be immutable if possible (I read 'possible' as not requiring excessive effort and not being inconvenient for consumption).  Is there a simple way to achieve immutability here?

LaunchDarkly/LaunchDarkly/LDClientVariation.swift
kinyoklion: Ideally yes, it should be the specific method.

LaunchDarkly/LaunchDarkly/Models/Hooks/Hook.swift
keelerm84: Yes. When we add a new method on this protocol, we will need to also provide an extension to that protocol with a default implementation.

LaunchDarkly/LaunchDarkly/Models/Hooks/Hook.swift
keelerm84: Dictionaries in Swift are passed by value (but lazily copied). So providing a dictionary to a function is already an immutable operation.

LaunchDarkly/LaunchDarkly/LDClientVariation.swift
keelerm84: Done!

kinyoklion (approved):

tanderson-ld (approved):

PR #379, opened 2024-05-06
feat: Enable gzip support for event payloads

keelerm84:


tanderson-ld (approved):
Small comment, the parameter is called body which is generic, but then the internal variable is called eventPayload which is more specific. This indicates an assumption about how this function is being used.  Consider not using as specific a name.  Maybe, `bodyToUse`?

keelerm84:
> Small comment, the parameter is called body which is generic, but then the internal variable is called eventPayload which is more specific. This indicates an assumption about how this function is being used. Consider not using as specific a name. Maybe, `bodyToUse`?

Done! I went with `httpBody` to match the property assignment.

LaunchDarkly.podspec
```diff
@@ -39,5 +39,8 @@ Pod::Spec.new do |ld|
 
   ld.subspec 'Core' do |es|
     es.dependency 'LDSwiftEventSource', '3.1.1'
+    es.ios.dependency 'GZIP', '1.3.2'
```
keelerm84, line 42: This gzip library doesn't support watchos. Which is probably reasonable since CPU cycles are going to be more precious on that device.

tanderson-ld (approved):

Issue #316, opened 2023-10-26
Different Feature Flag Caching Mechanisms


brianmichel:
**Is your feature request related to a problem? Please describe.**

We use a fork of this SDK to power feature flagging on our macOS application. While we understand this is explicitly named the `ios-client-sdk` it serves as a defacto macOS SDK since it drops in without any issue (perhaps an issue we should discuss). Due to flags getting cached in user defaults anyone on macOS can simply use the `defaults` command to read and write data to these values which can easily break an application. This is a fine solution for iOS since there's no access to the device where you could run said command, but isn't sufficient for macOS implementations (or perhaps in any other environment capable of running Swift code). 

**Describe the solution you'd like**
We'd like the ability to specify the caching storage mechanism, and even provide our own to meet the needs of different scenarios/environments. You can see a quick and dirty proof of concept that I built on our fork (apologies for being a little out of date) https://github.com/launchdarkly/ios-client-sdk/compare/v5...thebrowsercompany:ios-client-sdk:brian/caching-options about how the system could work. You could imagine other implementations that serialize this information to a file somewhere else, or an encrypted database of some sort. You get the idea.

**Describe alternatives you've considered**
I'm open to other ideas, however I don't know what other alternative solutions here there might be since there isn't any option today of providing alternative implementations to the caching mechanism .

**Additional context**
N/A

louis-launchdarkly:
Hello @brianmichel, thank you for reaching out about this use case. While we cannot promise a timeline, this seems like a useful enhancement to have. I put this into the backlog and will discuss it with the team.

Filed as _222822_.

brianmichel:
> Hello @brianmichel, thank you for reaching out about this use case. While we cannot promise a timeline, this seems like a useful enhancement to have. I put this into the backlog and will discuss it with the team.
> 
> 
> 
> Filed as _222822_.

Thanks for the quick reply @louis-launchdarkly! Do you think an approach similar to what I've linked to would be sufficient? I think I can make it a bit more dynamic so someone could easily pass in a custom instance as some sort of associated value on an enumeration.

I'd be happy to try to contribute that back if you think the design is sufficient.

brianmichel:
Hey @louis-launchdarkly just wanted to see if there was any update here before the end of the year

nalexn:
Hey @louis-launchdarkly ,

[...]

PR #412, opened 2024-10-25
feat: Add support for external cache storage

nalexn:
**Requirements**

- [x] I have followed the repository's [pull request submission guidelines](../blob/v9/CONTRIBUTING.md#submitting-pull-requests)

**Related issues**

Provide links to any issues in this repository or elsewhere relating to this pull request: #316
Also, provides an alternative to #408 when using `LDInMemoryCache`

This PR adds a new property to the `LDConfig` file allowing to replace the default cache storage (`NSUserDefaults`) with either `LDInMemoryCache`, `LDFileCache`, or a custom cache user implements by conforming to `KeyedValueCaching` protocol.

```swift
var config = LDConfig(mobileKey: ..., autoEnvAttributes: .disabled)
config.cacheFactory = LDFileCache.factory(encryptionKey: ...)
```

Provide a clear and concise description of any alternative solutions or features you've considered.

**Additional context**

* 100% backward compatible, all inner workings of the library don't change if the `LDConfig` isn't touched
* Both `LDInMemoryCache` and `LDFileCache` are thread-safe and fully covered with tests
* `LDFileCache` supports optional encryption of the file with the AES algorithm
* `LDFileCache` supports migration from the `NSUserDefaults` so there is no flag flicker when updating
* Writing to file happens from a background queue and is throttled

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -752,7 +752,7 @@ public class LDClient {
 
         os_log("%s LDClient starting", log: config.logger, type: .debug, typeName(and: #function))
 
-        let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger, cacheBuilder: config.cacheBuilder)
+        let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger, cacheFactory: config.cacheFactory)
         var keys = [config.mobileKey]
         keys.append(contentsOf: config.getSecondaryMobileKeys().values)
         serviceFactory.makeCacheConverter().convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -756,7 +756,7 @@ public class LDClient {
         var keys = [config.mobileKey]
         keys.append(contentsOf: config.getSecondaryMobileKeys().values)
         let cacheConverter = serviceFactory.makeCacheConverter()
-        cacheConverter.migrateStorage(serviceFactory: serviceFactory, keysToConvert: keys)
+        cacheConverter.migrateStorage(serviceFactory: serviceFactory, keysToMigrate: keys, from: LDConfig.Defaults.cacheFactory)
         cacheConverter.convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
         var mobileKeys = config.getSecondaryMobileKeys()
         var internalCount = 0
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -752,7 +752,7 @@ public class LDClient {
 
         os_log("%s LDClient starting", log: config.logger, type: .debug, typeName(and: #function))
 
-        let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger)
+        let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger, cacheBuilder: config.cacheBuilder)
         var keys = [config.mobileKey]
         keys.append(contentsOf: config.getSecondaryMobileKeys().values)
         serviceFactory.makeCacheConverter().convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -755,7 +755,9 @@ public class LDClient {
         let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger, cacheFactory: config.cacheFactory)
         var keys = [config.mobileKey]
         keys.append(contentsOf: config.getSecondaryMobileKeys().values)
-        serviceFactory.makeCacheConverter().convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
+        let cacheConverter = serviceFactory.makeCacheConverter()
+        cacheConverter.migrateStorage(serviceFactory: serviceFactory, keysToConvert: keys)
+        cacheConverter.convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
         var mobileKeys = config.getSecondaryMobileKeys()
         var internalCount = 0
         let completionCheck = {
--- a/LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
+++ b/LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
@@ -254,7 +254,7 @@ public struct LDConfig {
         static let logger: OSLog = OSLog(subsystem: "com.launchdarkly", category: "ios-client-sdk")
 
         /// The default cache for feature flags is UserDefaults
-        static let cacheFactory: CacheFactory = { _, cacheKey in
+        static let cacheFactory: CacheFactory = { cacheKey, _ in
             UserDefaults(suiteName: cacheKey)!
         }
 
--- a/LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
+++ b/LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
@@ -253,6 +253,11 @@ public struct LDConfig {
         /// The default logger for the SDK. Can be overridden to provide customization.
         static let logger: OSLog = OSLog(subsystem: "com.launchdarkly", category: "ios-client-sdk")
 
+        /// The default cache for feature flags is UserDefaults
+        static let cacheBuilder: (String?) -> KeyedValueCaching = { cacheKey in
+            UserDefaults(suiteName: cacheKey)!
+        }
+
         /// The default behavior for event payload compression.
         static let enableCompression: Bool = false
     }
@@ -426,6 +431,9 @@ public struct LDConfig {
 
     /// Configure the logger that will be used by the rest of the SDK.
     public var logger: OSLog = Defaults.logger
+
+    /// Configure the persistent storage for caching flags locally
+    public var cacheBuilder: (String?) -> KeyedValueCaching = Defaults.cacheBuilder
 
     /// LaunchDarkly defined minima for selected configurable items
     public let minima: Minima
--- a/LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
+++ b/LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
@@ -254,7 +254,7 @@ public struct LDConfig {
         static let logger: OSLog = OSLog(subsystem: "com.launchdarkly", category: "ios-client-sdk")
 
         /// The default cache for feature flags is UserDefaults
-        static let cacheBuilder: (String?) -> KeyedValueCaching = { cacheKey in
+        static let cacheFactory: CacheFactory = { _, cacheKey in
             UserDefaults(suiteName: cacheKey)!
         }
 
@@ -433,7 +433,7 @@ public struct LDConfig {
     public var logger: OSLog = Defaults.logger
 
     /// Configure the persistent storage for caching flags locally
-    public var cacheBuilder: (String?) -> KeyedValueCaching = Defaults.cacheBuilder
+    public var cacheFactory: CacheFactory = Defaults.cacheFactory
 
     /// LaunchDarkly defined minima for selected configurable items
     public let minima: Minima
--- a/LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
+++ b/LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
@@ -57,18 +57,24 @@ protocol FeatureFlagCaching {
     func saveCachedData(_ storedItems: StoredItems, cacheKey: String, contextHash: String, lastUpdated: Date, etag: String?)
 }
 
+extension MobileKey {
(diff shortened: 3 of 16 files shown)


tanderson-ld:
Hi @nalexn, thank you for your contribution.  We on the SDK team will take a look early next week and provide our thoughts and feedback.

I did take a brief look through just now.  My initial reaction is your code is thoughtful, so thank you for that. Some tasks popped into my head that I want to document to come back to during review:
- During review, examine guarantees around data making it to persistence.
- Discuss migration pros/cons.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
```diff
@@ -57,18 +57,24 @@ protocol FeatureFlagCaching {
...
+    func cacheKey() -> String {
+        let cacheKey: String
+        if let bundleId = Bundle.main.bundleIdentifier {
+            cacheKey = "\(Util.sha256base64(bundleId)).\(Util.sha256base64(self))"
+        } else {
+            cacheKey = Util.sha256base64(self)
+        }
+        return "com.launchdarkly.client.\(cacheKey)"
+    }
+}
```
tanderson-ld, line 70: Is this extension organizational or is there another reason for adding it?

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -752,10 +752,12 @@ public class LDClient {
...
         os_log("%s LDClient starting", log: config.logger, type: .debug, typeName(and: #function))
 
-        let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger)
+        let serviceFactory = serviceFactory ?? ClientServiceFactory(logger: config.logger, cacheFactory: config.cacheFactory)
         var keys = [config.mobileKey]
         keys.append(contentsOf: config.getSecondaryMobileKeys().values)
-        serviceFactory.makeCacheConverter().convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
+        let cacheConverter = serviceFactory.makeCacheConverter()
+        cacheConverter.migrateStorage(serviceFactory: serviceFactory, keysToMigrate: keys, from: LDConfig.Defaults.cacheFactory)
+        cacheConverter.convertCacheData(serviceFactory: serviceFactory, keysToConvert: keys, maxCachedContexts: config.maxCachedContexts)
```
keelerm84, line 760: With caching implementations being as wildly different as they are, it seem like the `convertCacheData` should be implemented alongside the cache implementation.

So I think the cache prototype probably needs a method to retrieve an optional converter. If provided, then we pass in the list of mobile keys, and the max cached contexts option. From there, the implementation is cache specific.

We should also conversion migration and conversion as the same concept. If someone wants to migrate between two completely different storage technologies, they can write a custom cacher to handle that. Seems like it would be straight forward enough.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
```diff
@@ -57,18 +57,24 @@ protocol FeatureFlagCaching {
     func saveCachedData(_ storedItems: StoredItems, cacheKey: String, contextHash: String, lastUpdated: Date, etag: String?)
 }
 
+extension MobileKey {
+    func cacheKey() -> String {
```
keelerm84, line 61: What do you think about having the cache define it's own key? I don't see any reason why completely different implementations need to use the same format as this one.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/LDFileCache.swift
```diff
@@ -0,0 +1,125 @@
+import Foundation
+import OSLog
+
+public final class LDFileCache: KeyedValueCaching {
```
keelerm84, line 4: I'm hesitant to include this in the SDK. While this implementation seems sufficient for now, over time it becomes another point of divergence we need to maintain and test. This is doubly true for an implementation that deals with encryption in any sort of way.

In general, I think if we can avoid assuming that encryption burden for our customers. This would be a good candidate for a standalone library that integrates with the LD SDK if you wanted to make your implementation available to the public.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/LDInMemoryCache.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+public final class LDInMemoryCache: KeyedValueCaching {
```
keelerm84, line 3: I'm okay with keeping the in memory cache implementation since the maintenance burden is greatly reduced. Also this would immediately prove useful for unit testing.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/KeyedValueCache.swift
```diff
@@ -1,25 +1,14 @@
...
-
-    func removeAll() {
-        dictionaryRepresentation().keys.forEach { removeObject(forKey: $0) }
-    }
-
-    func keys() -> [String] {
-        dictionaryRepresentation().keys.map { String($0) }
-    }
+public extension LDConfig {
+    typealias CacheFactory = (_ cacheKey: String?, _ logger: OSLog) -> KeyedValueCaching
```
keelerm84, line 13: Is there any reason to make the cache key optional? It forces a decision on what a default key would be in your cache implementations which I would prefer to avoid.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/LDInMemoryCache.swift
```diff
@@ -0,0 +1,52 @@
...
+    private static let instancesLock = NSLock()
+
+    private var cache: [String: Any] = [:]
+    private var cacheLock = NSLock()
+
+    public static func factory() -> LDConfig.CacheFactory {
+        return { cacheKey, _ in
+            instancesLock.lock()
+            defer { instancesLock.unlock() }
+            let cacheKey = cacheKey ?? "default"
```
keelerm84, line 15: As mentioned above, if the cache key isn't nillable, then we don't have to hard code this "default" which I would really rather avoid.

LaunchDarkly/LaunchDarkly/Util.swift
```diff
@@ -28,3 +80,28 @@ extension String {
         return true
     }
 }
+
+extension DispatchQueue {
+
+    func debouncer() -> Debouncer {
```
keelerm84, line 86: I can see this method being misused. It would seem like I could call `debouncer` on any queue from two different areas, and it would handle `debouncing` both of those requests. But that's not the case since it would be two instances with different locks.

LaunchDarkly/LaunchDarkly/Util.swift
```diff
@@ -28,3 +80,28 @@ extension String {
...
+        private let lock = NSLock()
+        private let queue: DispatchQueue
+        private var workItem: DispatchWorkItem?
+
+        fileprivate init(queue: DispatchQueue) {
+            self.queue = queue
+        }
+
+        func debounce(interval: DispatchTimeInterval, action: @escaping () -> Void) {
+            lock.lock(); defer { lock.unlock() }
```
keelerm84, line 100:
```suggestion
            lock.lock()
            defer { lock.unlock() }
```

LaunchDarkly/LaunchDarkly/Util.swift
```diff
@@ -17,6 +27,48 @@ class Util {
         }
         return Data(digest)
     }
+
+    class func encrypt(_ data: Data, encryptionKey: String, cacheKey: String) throws -> Data {
```
keelerm84, line 31: As mentioned above, this isn't something we are ready to bring into the SDK. This, along with the file cache, can be tucked into that same external library.

LaunchDarkly/LaunchDarkly/Models/LDConfig.swift
```diff
@@ -253,6 +253,11 @@ public struct LDConfig {
         /// The default logger for the SDK. Can be overridden to provide customization.
         static let logger: OSLog = OSLog(subsystem: "com.launchdarkly", category: "ios-client-sdk")
 
+        /// The default cache for feature flags is UserDefaults
+        static let cacheFactory: CacheFactory = { cacheKey, _ in
+            UserDefaults(suiteName: cacheKey)!
```
keelerm84, line 258: I realize we were previously forcing the unwrap here and you just moved this bit of code.

Now that we have an in memory cache that can't fail, what about having this fall back to that if this instantiation fails? Then we don't have to worry about the force unwrap anymore.

keelerm84 (changes requested):

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/KeyedValueCache.swift
nalexn: This was dictated by the `convertCacheData` function, which is migrating UserDefaults with key = nil

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/LDFileCache.swift
tanderson-ld: An additional consideration with the debouncer is there will be a change in the when data becomes durable, where that happens, and the probability of it happening.

With UserDefaults, it is my understanding that the API guarantees eventual durability except in the case of OS/phone battery failure.  The issue you linked to in the PR description where flicker is seen on relaunch is a result of the OS not sharing the view of in memory caching of UserDefaults across processes.  Definitely an issue, but not trivial to address without blocking logic somewhere.

With the debouncer, there is a window of time where if two updates come in barely before the application is killed, the second update will affect flag evaluations for a moment, but then may not make it to persistence.  This could have the same effect of "flicker" where the next app launch initializes with flag data from "the past", similar to the UserDefaults case.

Based on @keelerm84's comment above, if the debouncer goes with this file, this behavior would be in your control.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
nalexn: I didn't consider this as an important thing to add (trying to make the PR smaller), but this certainly can be done. Ideally, a wider refactoring is needed here. The parameter name `cacheKey` is confusing because it means different entities in different contexts, while always being a `String`. This cost me a couple of hours of painful debugging to realize that:

* `makeKeyedValueCache(cacheKey: String?)`
* `func getCachedData(cacheKey: String, ...)` and `func saveCachedData(_ storedItems: StoredItems, cacheKey: String, ...)`

use cacheKey that look different, and are generated differently. The first one usually looks like `com.launchdarkly.client....`, and the second one is `context.fullyQualifiedHashedKey()`

LaunchDarkly/LaunchDarkly/Util.swift
nalexn: Well, that's a `debouncer`, not a `debounce` method. If you don't retain it - the work item won't be executed at all. Obviously it can be misused, just like entire GCD. Happy to see a suggestion on alternative approach. Late enough I found a `throttle` in the codebase, maybe something that reuses it

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
nalexn: Code reuse

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/KeyedValueCache.swift
keelerm84: I think that's actually in line with my later comment about this conversion being something associated with a specific cache implementation. Our existing implementation could take its cache key, but when doing the conversion bit, it can ignore that key then since it knows it didn't use to use it.

Also, at this point, I think we could argue that our cache converter could probably be cleaned up some. I don't think we have to support a v5 -> v9 migration path. So we could probably remove those steps entirely. What do you think @tanderson-ld

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/FeatureFlagCache.swift
keelerm84: That's fair. It may be something we add ourselves then after this lands. 

@tanderson-ld that is something you and I can decide on then if we want this now, later, or not at all.

LaunchDarkly/LaunchDarkly/Util.swift
keelerm84: > Well, that's a debouncer, not a debounce method.

Yeah, I see that. I guess it is that it's an extension of the queue that makes me think it would be unique across usages. `DispatchQueue.main.debouncer().debounce(....)` seems like it would be kind of convenient usage, and would probably be something I'd want to attempt. 

We don't have to worry about it.

LaunchDarkly/LaunchDarkly/ServiceObjects/Cache/KeyedValueCache.swift
tanderson-ld: I agree.  Dropping support for converting v5 to remove the optionality of the cacheKey is a good tradeoff.

I think these are the pertinent lines.
```
        // Remove V5 cache data
        let standardDefaults = serviceFactory.makeKeyedValueCache(cacheKey: nil)
        standardDefaults.removeObject(forKey: "com.launchdarkly.dataManager.userEnvironments")
```

nalexn:
All the suggested improvements and refactorings totally make sense, but my team is limited on the time we can dedicate to solving our peculiar problem, so we'll likely just use our branch for the time being. The PR can be closed or updated at your discretion.

keelerm84:
> All the suggested improvements and refactorings totally make sense, but my team is limited on the time we can dedicate to solving our peculiar problem, so we'll likely just use our branch for the time being. The PR can be closed or updated at your discretion.

Thank you for letting me know. I'll take a look at folding in those changes for you.

PR #414, opened 2024-11-06
feat: Add `LDConfig.sendEvents` option to disable all events

keelerm84:


LaunchDarkly/LaunchDarkly/Networking/DarklyService.swift
```diff
@@ -60,7 +60,7 @@ final class DarklyService: DarklyServiceProvider {
         self.context = context
         self.serviceFactory = serviceFactory
 
-        if !config.mobileKey.isEmpty && !config.diagnosticOptOut {
+        if !config.mobileKey.isEmpty && !config.diagnosticOptOut && config.sendEvents {
```
keelerm84, line 63: This change alone is sufficient to disable diagnostic events. If the cache doesn't exist, we won't collect diagnostic events.

LaunchDarkly/LaunchDarkly/ServiceObjects/ClientServiceFactory.swift
```diff
@@ -121,8 +125,12 @@ final class ClientServiceFactory: ClientServiceCreating {
         DiagnosticCache(sdkKey: sdkKey)
     }
 
-    func makeDiagnosticReporter(service: DarklyServiceProvider, environmentReporter: EnvironmentReporting) -> DiagnosticReporting {
-        DiagnosticReporter(service: service, environmentReporting: environmentReporter)
+    func makeDiagnosticReporter(config: LDConfig, service: DarklyServiceProvider, environmentReporter: EnvironmentReporting) -> DiagnosticReporting {
+        if config.sendEvents && !config.diagnosticOptOut {
```
keelerm84, line 129: But we also add this bit to make it a little more efficient by not making a reporter at all.

kinyoklion: The suspense between the two comments. I read the first one and I was like, "yes, but that indirectness scares me".

keelerm84: 😆 yes, I hate that there are even two paths like this. But there are protocols that are depending on others in a cycle and so really teasing that out as much as I would prefer is probably more effort than it is worth. Especially since we're pushing for a larger re-work.

kinyoklion (approved):

PR #422, opened 2025-01-13
Relying on CustomStringConvertible for description of FlagSyncResult

rkreutz:
Currently the SDK relies on performing flag sync up post-processing on the Main Thread, this in turn triggers `LDClient.onFlagSyncComplete(result:)` which has some `os_log` calls where the result is transformed into a `String` using `String.init(describing:)`, this in turn uses Swift default implementation on stringifying objects which for large collections can be a big bottle-neck, we could easily identify instances where the Main Thread was running this piece of code for 200ms (we have around 280 flags in our project), causing hitches/hangs in our project.

To address the issue I've conformed `FlagSyncResult` to `CustomStringConvertible` and provided a custom description, it mostly behaves the same with the exception for `.flagCollection` case in which we use `JSONEncoder` to stringify the list of flags.

We've also transformed `StoredItems` type alias into a struct wrapping the dictionary in order to be able to provide a similar `CustomStringConvertible` conformance.

With this change we could notice the execution time of `LDClient.onFlagSyncComplete(result:)` has dropped to levels between 30-40% of the original execution time.

Bellow you can notice some stack traces while profiling our application along with a heat map of the methods being called by  `LDClient.onFlagSyncComplete(result:)`

|Before|After|
|---|---|
|Screenshot 2025-01-13 at 12 44 17|Screenshot 2025-01-13 at 12 39 19|
|Screenshot 2025-01-13 at 12 44 41|Screenshot 2025-01-13 at 12 39 41|

**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [ ] I have followed the repository's [pull request submission guidelines](../blob/v9/CONTRIBUTING.md#submitting-pull-requests)

**Related issues**

https://support.launchdarkly.com/hc/en-us/requests/84426

Providing `CustomStringConvertible` conformance to relevant components.

Ideally flag sync-up post-processing wouldn't be done on the Main Thread, but I'm uncertain what implications this could have so I've opted for the current solution.

**Additional context**

N/A

### DIFF
--- a/LaunchDarkly/LaunchDarkly/ServiceObjects/FlagStore.swift
+++ b/LaunchDarkly/LaunchDarkly/ServiceObjects/FlagStore.swift
@@ -46,17 +46,59 @@ enum StorageItem: Codable {
     }
 }
 
-typealias StoredItems = [LDFlagKey: StorageItem]
-extension StoredItems {
+struct StoredItems {
+    var items: [LDFlagKey: StorageItem]
+
     var featureFlags: [LDFlagKey: FeatureFlag] {
-        self.compactMapValues {
+        items.compactMapValues {
             guard case .item(let flag) = $0 else { return nil }
             return flag
         }
     }
 
+    var isEmpty: Bool {
+        items.isEmpty
+    }
+
+    subscript (key: LDFlagKey) -> StorageItem? {
+        get { items[key] }
+        set { items[key] = newValue }
+    }
+
+    mutating func updateValue(_ newValue: StorageItem, forKey key: LDFlagKey) {
+        items.updateValue(newValue, forKey: key)
+    }
+}
+
+extension StoredItems {
     init(items: [LDFlagKey: FeatureFlag]) {
-        self = items.mapValues { .item($0) }
+        self.items = items.mapValues { .item($0) }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


tanderson-ld:
Hi @rkreutz.  Thank you for bringing this to our attention and taking the time to make the contribution.  We will take a look.

tanderson-ld:
@rkreutz, most of our other SDKs check the logging level before performing string interpolation / serialization.  The serialization in this code path is not intended when not in debug.  We're inclined to just remove this log and replace it with summarized information such as "Flag result received with N flags." in the short term.  Does this meet your needs?

rkreutz:
Yeah removing os_logs here should work as well, I didn't go that route since I did not know whether other consumers relied on it, but if you are confident about removing those logs it should be fine as well

tanderson-ld:
#425 has merged and should address this issue.  We will be releasing that shortly.  Feel free to leave this open for the time being until you're able to test the fix so we can confirm it is resolved.

tanderson-ld:
Version 9.12.2 has been released.  Let us know if that fixes the performance issue for you.

rkreutz:
It did the trick, Main Thread usage is much lower now, thanks.

PR #425, opened 2025-01-15
fix: replaced inefficient string logic for sync complete log

tanderson-ld:
**Additional context**
String creation of flag collection was happening regardless of logging level.  This fix resolves the performance issue in a straightforward way while maintaining enough information to verify SDK is receiving flag data when debugging.

### DIFF
--- a/LaunchDarkly/LaunchDarkly/ServiceObjects/FlagChangeNotifier.swift
+++ b/LaunchDarkly/LaunchDarkly/ServiceObjects/FlagChangeNotifier.swift
@@ -27,23 +27,23 @@ final class FlagChangeNotifier: FlagChangeNotifying {
     }
 
     func addFlagChangeObserver(_ observer: FlagChangeObserver) {
-        os_log("%s observer: %s", log: logger, type: .debug, typeName(and: #function), String(describing: observer))
+        os_log("%s called for keys %s", log: logger, type: .debug, typeName(and: #function), String(describing: observer.flagKeys))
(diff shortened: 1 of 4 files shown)


keelerm84 (approved):

LaunchDarkly/LaunchDarkly/ServiceObjects/FlagChangeNotifier.swift
```diff
@@ -27,23 +27,23 @@ final class FlagChangeNotifier: FlagChangeNotifying {
...
 
     func addFlagChangeObserver(_ observer: FlagChangeObserver) {
-        os_log("%s observer: %s", log: logger, type: .debug, typeName(and: #function), String(describing: observer))
+        os_log("%s called for keys %s", log: logger, type: .debug, typeName(and: #function), String(describing: observer.flagKeys))
         flagChangeQueue.sync { flagChangeObservers.append(observer) }
     }
 
     func addFlagsUnchangedObserver(_ observer: FlagsUnchangedObserver) {
-        os_log("%s observer: %s", log: logger, type: .debug, typeName(and: #function), String(describing: observer))
+        os_log("%s called.", log: logger, type: .debug, typeName(and: #function), String(describing: observer))
```
keelerm84:
```suggestion
        os_log("%s called.", log: logger, type: .debug, typeName(and: #function))
```

keelerm84 (commented):
Aside from that one comment it looks good to me.

PR #429, opened 2025-02-12
fix: fixes race condition that could lead to multiple start completion invocations

tanderson-ld:


keelerm84 (approved):

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -827,24 +827,25 @@ public class LDClient {
...
-            completion?(completed)
+            completion?(true) // offline is considered a short circuited timed out case
         } else {
             let startTime = Date().timeIntervalSince1970
             start(serviceFactory: serviceFactory, config: config, context: context) {
                 internalCompletedQueue.async {
-                    if startTime + startWaitSeconds > Date().timeIntervalSince1970 && completed {
-                        completed = false
-                        completion?(completed)
+                    if startTime + startWaitSeconds > Date().timeIntervalSince1970 && !completed {
```
macdrevx, line 839: I think you could just remove the `startTime + startWaitSeconds > Date().timeIntervalSince1970` condition. If the `asyncAfter` closure runs first, then you know you timed out and `completed` will be true

macdrevx: And if this closure runs before the one passed to `asyncAfter` but `startTime + startWaitSeconds > Date().timeIntervalSince1970` is false (e.g. because of the leeway that dispatch allows with `asyncAfter`) it seems unnecessary to fail just due to time if you know you've actually succeeded.

PR #451, opened 2025-09-11
chore!: bump target versions to iOS13 and others

abelonogov-ld:
Bumps supported versions of platforms:
- iOS "12.0" -> "13.0"
- watchOS "4.0" -> "6.0"
- tvOS "12.0" -> "13.0"
- masOS "10.13" -> "10.15"

**Requirements**

- [x] I have followed the repository's [pull request submission guidelines](../blob/v9/CONTRIBUTING.md#submitting-pull-requests)
except I launch unit tests using "swift test" due not having Apple Developer Account

**Related issues**

OpenTelemetry specifies iOS13 in their package file
https://github.com/open-telemetry/opentelemetry-swift/blob/main/Package.swift

Alternative would be using feature branch or forking and refactoring OpenTelemetry to iOS12

**Additional context**

Observability teams uses OpenTelemetry has lowest target iOS13. We need to bump platform versions in LD SDK. To be able build Observability library.
As far we know we don't have customers on iOS12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to the LaunchDarkly iOS SDK will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org).
 
+## [9.16.0](https://github.com/launchdarkly/ios-client-sdk/compare/9.15.0...9.16.0) (2025-09-11)
+
+
+### Features
+* bumps supported versions of platforms:
+- iOS "12.0" -> "13.0"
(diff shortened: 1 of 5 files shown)


tanderson-ld:
These changes are going into the v9 branch, but the breaking-ness will lead to v10 being released. It may make sense to make a v10 branch if the intention is to release v10.

There is an epic containing breaking work, consider reviewing that epic and doing the low hanging fruit stories such as removing deprecated methods.

tanderson-ld (approved):

PR #487, opened 2026-03-19
fix: Call identify hooks during init.

beekld:
**Requirements**

- [X] I have followed the repository's [pull request submission guidelines](../blob/v11/CONTRIBUTING.md#submitting-pull-requests)

**Related issues**

N/A

Before this change, `beforeIdentify` and `afterIdentify` were only called when `identify` was called. After this change, they are also called as part of `init`.

None.

**Additional context**

I don't think this would be considered a breaking change, but I wouldn't mind a second opinion. The change in behavior is observable by the user, but the new behavior is what was originally intended. So I would consider this a bugfix, and would not expect customers to need to make any changes to their code to deal with the change.

---

> [!NOTE]
> **Medium Risk**
> Changes observable hook execution timing by invoking `beforeIdentify`/`afterIdentify` during `LDClient` init and by including plugin-provided hooks in that lifecycle, which could affect apps with side-effecting hooks. Scope is limited to hook plumbing and tests, with no auth or data model changes.
> 
> **Overview**
> `LDClient` now executes identify hooks as part of initialization: it runs `beforeIdentify` with method name `"init"` and defers `afterIdentify` until the initial `setOnline` completes.
> 
> Plugin hooks are collected earlier (during `LDClient` init rather than `LDClient.start`) so plugin-provided hooks participate in the init identify lifecycle, and the hook helper APIs in `LDClientIdentifyHook.swift` are widened to `internal` to support this.
> 
> Tests are updated/expanded to assert the additional init-time hook calls, preserve hook ordering, and verify plugin hooks fire during init.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ab0372abbf967254937777120a8234bd20548ab3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -970,10 +970,8 @@ public class LDClient {
 
         var hookState: IdentifyHookState? = nil
         if !hooks.isEmpty {
-            let seriesContext = IdentifySeriesContext(context: context, methodName: "identify")
-            let seriesData = hooks.map { hook in
-                hook.beforeIdentify(seriesContext: seriesContext, seriesData: EvaluationSeriesData())
-            }
+            let seriesContext = IdentifySeriesContext(context: context, methodName: "init")
+            let seriesData = hooks.map { hook in hook.beforeIdentify(seriesContext: seriesContext, seriesData: EvaluationSeriesData()) }
             hookState = IdentifyHookState(seriesContext: seriesContext, seriesData: seriesData, hooksSnapshot: hooks)
         }
 
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -995,12 +995,16 @@ public class LDClient {
             flagStore.replaceStore(newStoredItems: cachedFlags)
         }
 
+        let hookState = executeBeforeIdentifyHooks(context: context)
         eventReporter.record(IdentifyEvent(context: context))
         self.connectionInformation = ConnectionInformation.uncacheConnectionInformation(config: config, ldClient: self, clientServiceFactory: self.serviceFactory)
 
         internalSetOnline(configuration.startOnline) {
             os_log("%s LDClient started", log: configuration.logger, type: .debug, self.typeName(and: #function))
             completion?()
+            if let state = hookState {
+                self.executeAfterIdentifyHooks(state: state, result: .complete)
+            }
         }
     }
 }
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -836,17 +836,6 @@ public class LDClient {
                 credential: mobileKey
             )
 
-            // add all the plugin hooks
-            for plugin in config.plugins {
-                // Catch to protect against any runtime exceptions from plugin
-                do {
-                    let pluginHooks = try plugin.getHooks(metadata: environmentMetadata)
-                    instance.hooks.append(contentsOf: pluginHooks)
-                } catch {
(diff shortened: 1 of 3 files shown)


LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -995,12 +995,16 @@ public class LDClient {
...
+        let hookState = executeBeforeIdentifyHooks(context: context)
         eventReporter.record(IdentifyEvent(context: context))
         self.connectionInformation = ConnectionInformation.uncacheConnectionInformation(config: config, ldClient: self, clientServiceFactory: self.serviceFactory)
 
         internalSetOnline(configuration.startOnline) {
             os_log("%s LDClient started", log: configuration.logger, type: .debug, self.typeName(and: #function))
             completion?()
+            if let state = hookState {
+                self.executeAfterIdentifyHooks(state: state, result: .complete)
+            }
```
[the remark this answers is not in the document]
beekld: I believe this is a real issue, and I'm looking into it.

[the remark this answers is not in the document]
beekld: Okay, I _think_ I have fixed this by collecting all hooks in `init`, rather than waiting until `start` to collect them. It looks like that is what the Flutter SDK is doing. And I don't see any reason why it wouldn't work. But I admit I'm new to these SDKs and don't know all the ins-and-outs of the lifecycle.

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -995,12 +1002,16 @@ public class LDClient {
             flagStore.replaceStore(newStoredItems: cachedFlags)
         }
 
+        let hookState = executeBeforeIdentifyHooks(context: context)
```
tanderson-ld: The beforeIdentify hooks should run after the auto environment attributes are added to the context (so that the implementer of the beforeIdentify call can see all the added attributes), and before we use that modified context to load cache or make network connections.

https://github.com/launchdarkly/sdk-specs/tree/main/specs/HOOK-hooks#conditional-requirement-1412

So likely it needs to be called right after this block, but double check.

```
        if config.autoEnvAttributes {
            context = AutoEnvContextModifier(environmentReporter: environmentReporter, logger: config.logger).modifyContext(context)
        }
```

You can think of "identify" as updating the context the SDK is going to use and do things with.  If the SDK does anything with the context that can be detected or observed by the outside world, such as making network requests or loading cache which impacts flag evals, and it does so before `beforeIdentify`, it has sorta violated the logical order and it is a externally detectable contradiction of the API.

tanderson-ld (changes requested):

LaunchDarkly/LaunchDarkly/LDClient.swift
beekld: Okay, I think I've got it, and I think that's the right place. I mean, it seems clear from the spec that `beforeIdentify` should be called before `flagCache.getCachedData` at least. It's not obvious on first glance whether the various `make` functions on the previous lines have side effects, so it's better to err on calling the hooks before them.

There was one slight complication -- Swift really does not like `init` methods calling member functions before all object properties have been initialized, and `flagSynchronizer` is initialized after the cache is loaded. But the code in `executeBeforeIdentifyHooks` is very short and doesn't access any properties other than `hooks`, so I copy-pasted the code into `init`.

Please take another look.

tanderson-ld: Did the commit get pushed?

beekld: 🤦‍♀️ no. okay. it's been pushed now. sorry about that.

tanderson-ld (approved):

PR #516, opened 2026-08-04
feat: add configurable evaluation exposure deduplication

abelonogov-ld:
image

**Summary**

Opt-in deduplication of evaluation exposures for hooks. A hook is told about every evaluation until you wrap it, so nothing changes for existing hooks:

```swift
config.hooks = [
    MetricsHook(),                                      // every evaluation
    DedupingHook(ObservabilityHook()),                  // 10 minute window
    DedupingHook(TelemetryHook(), window: 60),
    DedupingHook(ExperimentHook(), deduper: myCustomDeduper)
]
```

New public API:

- `HookDecorator` — an open class conforming to `Hook` that forwards every stage to the hook it wraps, and reports that hook's metadata. Decorators stack in either order.
- `DedupingHook` — a decorator that suppresses an evaluation whose result the wrapped hook has just been told about. Wrap with a window, or with a deduper of your own.
- `EvaluationExposureDeduper` — the policy. The window is its only setting. Subclass it to decide differently; `shouldRecord(key:now:)` and `reset()` are the only members `DedupingHook` calls.
- `EvaluationExposureKey` — what identifies an evaluation result: environment name, flag key, variation, flag version, experiment status, and fully qualified context key.
- `EvaluationSeriesContext.evaluationExposureKey` — resolves that identity on demand, so an evaluation costs a flag lookup only when a hook asks for one.

The policy keeps one record per flag per environment, holding the result that flag last reported. The wrapped hook hears about the flag again as soon as the result changes, and once per window while it stays the same. Tracking the last result rather than every result seen means a flag flipping back and forth cannot hide its flips, and it bounds the records to the flags the environment serves.

The decision is taken in `beforeEvaluation`, before the series opens, so a suppressed evaluation reaches neither stage of the wrapped hook. Hooks pair their stages: the observability plugin starts a span in the before stage and ends it in the after one, so suppressing only the after stage would leave that span in its map to be evicted later and exported with a meaningless duration and no `feature_flag` event.

`LDClient.identify` clears what the wrapped hook has been told about, so the first evaluation of each flag afterwards always reaches it. Analytics events are untouched: feature, debug, and summary events are still recorded for every evaluation, so the evaluation counts LaunchDarkly reports for a flag do not change.

- **A deduper declared on the `Hook` protocol.** It was the first shape this took. It puts a policy on the protocol that every hook implementer has to think about, and it cannot be composed, so a hook that wanted deduplication plus anything else had nowhere to put the second behavior. The decorator gives both without touching `Hook`.
- **Filtering only `afterEvaluation`.** This is what the browser and React Native observability plugins do, and it is much less machinery: no exposure key before the evaluation, no resolver, no suppression marker in the series data. It is wrong here, because the mobile observability hook pairs its stages, as above.
- **Deduplicating on every distinct exposure seen, rather than one record per flag.** A flag that flips between two results would report neither flip after the first, and the set of remembered exposures grows with every result a flag has ever had.
- **`Date()` for the window.** A correction that moves the device clock backwards leaves every recorded time in the future, so those flags stay suppressed until real time catches up. Windows are measured against `CLOCK_MONOTONIC_RAW`, exposed as `EvaluationExposureDeduper.monotonicNow()`, which no correction reaches and which, unlike `mach_absolute_time` and everything built on it, keeps counting while the device sleeps.

**Additional context**

- Additive and off by default. An unwrapped hook behaves exactly as it does today.
- A hook set on `LDConfig` is one instance shared by the clients for every environment in `secondaryMobileKeys`, and so is its deduper. The environment is therefore part of both the exposure key and the per-flag record; sharing a record across environments would make each look like the other having changed its result, and neither would ever be suppressed.
- Experiment status is a component of the key in its own right, because `versionForEvents` prefers the flag's own version: a prerequisite flipping can move an evaluation into or out of an experiment while it lands on the same variation of the same flag version.
- Mirrored on Android in launchdarkly/android-client-sdk, with docs in sdk-meta and ld-docs-private. Demonstrated in launchdarkly/hello-ios.

---

> [!NOTE]
> **Overview**
> Adds **opt-in deduplication** for hook evaluation callbacks: wrap a hook in `DedupingHook` (default 10‑minute window, or a custom `EvaluationExposureDeduper`) so repeated evaluations with the same `EvaluationExposureKey` skip both `beforeEvaluation` and `afterEvaluation` until the result changes, the window expires, or `identify` clears state.
> 
> Introduces **`HookDecorator`** for stacking hook behavior, **`EvaluationExposureKey`** / **`evaluationExposureKey`** on `EvaluationSeriesContext`, and **`UnfairLock`** for thread-safe dedupe state. **`LDClient`** now hashes the mobile key once (`mobileKeyHash`), collects hooks deterministically at init, and reads each flag once per variation so hooks and the returned value describe the same snapshot.
> 
> Unwrapped hooks and LaunchDarkly **feature/debug/summary events** still run on every evaluation; only wrapped hook stages are suppressed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7b1404a17413bf770b4ead1e086da4b10bf4802. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


abelonogov-ld:
@launchdarkly verify changes here with Android https://github.com/launchdarkly/android-client-sdk main

abelonogov-ld:
@devin verify changes here with Android https://github.com/launchdarkly/android-client-sdk main

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -276,8 +276,19 @@ public class LDClient {
...
+    /// mobile key in the configuration, so it is known before the client exists rather than being learned as the client
+    /// starts: a hook that runs while the client is initializing is told which environment it belongs to.
+    let environmentName: String
+    /// Identifies this client's environment to a hook, without handing it the mobile key that identifies the
+    /// environment to LaunchDarkly. Hashed once here rather than per evaluation, because a deduping hook asks for it on
+    /// a path an application may take on every redraw of a view.
+    let mobileKeyHash: String
     let service: DarklyServiceProvider
-    var hooks: [Hook]
+    /// The hooks registered with this client: the configuration's, followed by the ones its plugins contribute.
```
tanderson-ld:
```suggestion
    /// The hooks registered with this client: the configuration's.
```

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -276,8 +276,19 @@ public class LDClient {
     }
 
     let config: LDConfig
+    /// The name this client is registered under in `LDClient.instances`, which identifies its environment. It names a
+    /// mobile key in the configuration, so it is known before the client exists rather than being learned as the client
+    /// starts: a hook that runs while the client is initializing is told which environment it belongs to.
+    let environmentName: String
```
tanderson-ld: Dead property, not used.

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
```diff
@@ -0,0 +1,232 @@
...
+/// look like the other having changed its result, and neither would ever be suppressed.
+private struct TrackedFlag: Hashable {
+    let mobileKeyHash: String
+    let flagKey: LDFlagKey
+}
+
+/// The result a flag last reported, and when.
+private struct LastReported {
+    let key: EvaluationExposureKey
+    let reportedAt: TimeInterval
```
tanderson-ld: A time interval for a reportedAt does not seem as correct as a type describing a point in time.  Can you just use the Uint64 of nanoseconds?  Then you create the TimeInterval when comparing and check if that time interval is less / greater than the window time interval.

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
```diff
@@ -0,0 +1,232 @@
...
+     clock cannot stretch a window: were this `Date()`, a correction that moved the clock backwards would leave every
+     recorded time in the future and suppress those flags until real time caught up. It also advances while the device
+     sleeps, unlike `mach_absolute_time` and everything built on it, such as `DispatchTime.now()` and
+     `ProcessInfo.systemUptime`, so a window is an interval of real time rather than of awake time.
+
+     Only differences between readings are meaningful: this is not a time of day, and comparing it with
+     `Date().timeIntervalSince1970` is a mistake.
+     */
+    public static func monotonicNow() -> TimeInterval {
+        return TimeInterval(clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)) / TimeInterval(NSEC_PER_SEC)
```
tanderson-ld: clock_gettime_nsec_np is only available on Darwin.  We are in process of a swift-client rewrite as well that we plan to use on Linxu and Windows if possible.  Are there any other options for getting uptime that meet the needs?

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
```diff
@@ -0,0 +1,232 @@
...
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(mobileKeyHash)
+        hasher.combine(flagKey)
+        hash(value: value, into: &hasher)
+        hasher.combine(variation)
+        hasher.combine(flagVersion)
+        hasher.combine(fullyQualifiedContextKey)
+    }
+
+    private func hash(value: LDValue, into hasher: inout Hasher) {
```
tanderson-ld: Was this auto generated?

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
```diff
@@ -0,0 +1,232 @@
...
+     Only differences between readings are meaningful: this is not a time of day, and comparing it with
+     `Date().timeIntervalSince1970` is a mistake.
+     */
+    public static func monotonicNow() -> TimeInterval {
+        return TimeInterval(clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)) / TimeInterval(NSEC_PER_SEC)
+    }
+
+    private let window: TimeInterval
+
+    private let queue = DispatchQueue(label: "com.launchdarkly.evaluationExposureDedupeQueue")
```
tanderson-ld: Consider NSLock for better performance and smaller memory footprint.  Also consider a RWLock since reads should be prioritized over writes especially in the rapid evaluation case.

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
```diff
@@ -0,0 +1,232 @@
...
+     DedupingHook(TelemetryHook(), window: 30),
+     DedupingHook(ExperimentHook(), deduper: myCustomDeduper)
+ ]
+ ```
+
+ This class is the SDK's implementation: it remembers the result each flag last reported, and tells the hook about the
+ flag again as soon as that result changes, or once the window elapses while it stays the same. Tracking one result per
+ flag rather than every result seen keeps a flag that flips back and forth from hiding the flips, and holds one record per
+ flag the application evaluates, so the window is the only thing there is to configure. Subclass this to implement a
+ different policy; only `shouldRecord(key:now:)` and `reset()` are called by `DedupingHook`.
```
tanderson-ld: I don't think recommending subclassing yet is a good idea.

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
tanderson-ld, line 1: Is the service objects folder the proper home for this?

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
```diff
@@ -0,0 +1,232 @@
...
+        queue.sync { lastReported.removeAll() }
+    }
+}
+
+/// The flag a record belongs to. The environment is part of it because a hook set on `LDConfig` is one instance shared
+/// by the clients for every environment in `secondaryMobileKeys`: were the environments to share a record, each would
+/// look like the other having changed its result, and neither would ever be suppressed.
+private struct TrackedFlag: Hashable {
+    let mobileKeyHash: String
+    let flagKey: LDFlagKey
```
tanderson-ld: Why does Android have an additional hash code, but this does not?

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
tanderson-ld, line 1: All public APIs related to this feature need to be marked with that experimental warning we discussed.  Especially important if customers will be customizing which properties they will be caring about in the deduping subclasses as that is likely to be a breaking point if we revise this in the future.

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
abelonogov-ld: It is optimization

abelonogov-ld: LDValue wasn't Hashable

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
abelonogov-ld: Agree as I still believe the best course of action to make an option not too flexible API

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
abelonogov-ld: LDValue wasn't hashable

LaunchDarkly/LaunchDarkly/ServiceObjects/EvaluationExposureDeduper.swift
abelonogov-ld: TimeInterval indicates that is Double encoding seconds, Apple has methods like timeintervalSince1970 and Apple consider any point in time as interval from some point.

tanderson-ld (approved):

PR #522, opened 2026-08-21
feat: add registerPlugin so a plugin can be added after the client starts

abelonogov-ld:
## Summary

Plugins could only be supplied through `LDConfig`, so an integration that learns about a plugin later — or that wants to instrument a client it did not configure — had no way in. This adds `LDClient.registerPlugin(_:)`, matching the method the .NET and Flutter SDKs already expose and the one being added to the Android SDK in launchdarkly/android-client-sdk#393.

- **Hooks became mutable, because registration after start cannot extend a constant array.** They now live in a `private var` behind an `NSLock` and are replaced rather than mutated in place. Each series reads a snapshot once into a local, so the hooks a series ends with are the hooks it began with — were the array read again mid-series, a hook registered in between would be handed an `after` stage for a series whose `before` stage it was never in. The three read sites (`evaluateWithHooks`, `executeBeforeIdentifyHooks`, `executeAfterTrackHooks`) all take that snapshot up front.
- **Hooks go live only once `register` returns.** This matches the Android and .NET ordering, and differs from configuration-time registration where a plugin's hooks are active before `register` is called. So a plugin's own hooks will not observe evaluations or identify calls its `register` makes; everything afterwards does run them.
- **`EnvironmentMetadata` is retained on the instance**, so a plugin registered later is handed the same environment description as one configured up front. This also removes the duplicate construction that previously existed in both `start` and `collectHooks`.

Registration applies to the one client it is called on, so a multi-environment setup means calling it per environment.

Note that the iOS `Plugin` protocol has no `onPluginsReady`, so unlike Android this path is just `getHooks` then `register` then activate.

## Test plan

Five cases added to `LDClientPluginsSpec`, all passing on an iPhone 16e simulator alongside the two pre-existing plugin tests (7 total):

- [x] `testRegisterPluginPassesClientAndEnvironmentMetadata`
- [x] `testRegisterPluginActivatesBundledHooks`
- [x] `testRegisterPluginDoesNotRunTheRegisteringPluginsOwnHooks`
- [x] `testRegisterPluginHooksRunAfterConfiguredHooks`
- [x] `testRegisterPluginAppliesOnlyToTheClientItIsCalledOn`
- [x] SwiftLint introduces no new violations in the touched files

---

> [!NOTE]
> **Overview**
> Adds **`LDClient.registerPlugin(_:)`** so integrations can attach plugins after `start`, aligned with other LaunchDarkly mobile SDKs. Plugins still use `LDConfig.plugins` at startup; this path is for late discovery or clients the app did not configure.
> 
> **Hook list is now mutable and thread-safe.** Hooks move from a fixed `let` array to `storedHooks` guarded by `NSLock`, with `addHooks` appending plugin hooks at registration. Evaluation, identify, and track paths each **snapshot** `hooks` once per series so mid-flight `registerPlugin` cannot pair an `after` stage with a `before` the new hook never ran.
> 
> **`environmentMetadata` is stored on the client** and passed to `register` and `getHooks`, so late plugins get the same environment description as config-time plugins and duplicate metadata construction in `start` / `collectHooks` is removed.
> 
> `registerPlugin` calls `getHooks` and activates hooks **before** `Plugin.register`, matching config-time ordering (including work done inside `register`). Registration applies only to the client instance you call it on (multi-environment setups need per-client calls).
> 
> Tests in `LDClientPluginsSpec` cover metadata handoff, hook activation, ordering after config hooks, hooks during `register`, and per-environment isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c5db77049a69f0198d12726e4d61ce9203fa361. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -774,10 +774,9 @@ public class LDClient {
      Registers a single plugin with this client after the client has been configured and started. To register plugins
      before the client starts, set `LDConfig.plugins` instead.
 
-     The plugin's hooks are collected first, then `Plugin.register` is called, and only then do those hooks begin
-     running. This ordering differs from registration at configuration time, where a plugin's hooks are active before
-     `register` is called: here a plugin's own hooks will not observe flag evaluations or identify calls that its
-     `register` makes. Once this method returns, later evaluations, identify calls, and track calls do run them.
+     The plugin's hooks begin running before `Plugin.register` is called, as they do for a plugin registered at
+     configuration time, so a plugin's own hooks observe the flag evaluations and identify calls that its `register`
+     makes. A plugin whose `register` fails keeps contributing its hooks, again matching configuration time.
 
      This registers the plugin with this client only. In a multi-environment configuration each environment has its own
      client, so registering with every environment means calling this on each of them.
@@ -785,9 +784,8 @@ public class LDClient {
      - parameter plugin: The plugin to register.
      */
     public func registerPlugin(_ plugin: Plugin) {
-        let pluginHooks = plugin.getHooks(metadata: environmentMetadata)
+        addHooks(plugin.getHooks(metadata: environmentMetadata))
         plugin.register(client: self, metadata: environmentMetadata)
-        addHooks(pluginHooks)
     }
 
     /**
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -281,10 +281,34 @@ public class LDClient {
     /// a path an application may take on every redraw of a view.
     let mobileKeyHash: String
     let service: DarklyServiceProvider
-    /// The hooks registered with this client: the configuration's.
-    /// Constant, so that a series reading it more than once, as an evaluation series does for its before and after
-    /// stages, runs the same hooks in both, whichever thread the evaluation was made from.
-    let hooks: [Hook]
+    /// The hooks registered with this client: the configuration's, then those contributed by plugins.
+    /// Replaced rather than modified in place, so that a caller of `registerPlugin` on one thread cannot be seen half
+    /// way by a series running on another.
+    private var storedHooks: [Hook]
+    private let hooksLock = NSLock()
+
+    /// A snapshot of the hooks to run. Every series reads this once into a local and works from that, so the hooks a
+    /// series ends with are the hooks it began with: were it read again mid-series, a hook registered in between would
+    /// be given an "after" stage for a series whose "before" stage it was never in.
+    var hooks: [Hook] {
+        hooksLock.lock()
+        defer { hooksLock.unlock() }
+        return storedHooks
+    }
+
+    /// Describes this client's environment to a plugin. Retained so that a plugin registered later, via
+    /// `registerPlugin`, is handed the same description as one configured up front.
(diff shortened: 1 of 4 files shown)


LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -281,10 +281,34 @@ public class LDClient {
...
+    /// Describes this client's environment to a plugin. Retained so that a plugin registered later, via
+    /// `registerPlugin`, is handed the same description as one configured up front.
+    let environmentMetadata: EnvironmentMetadata
+
+    /// Adds hooks, which the next series to begin will run. A series already under way runs the hooks it began with.
+    private func addHooks(_ newHooks: [Hook]) {
+        guard !newHooks.isEmpty else { return }
+
+        hooksLock.lock()
+        defer { hooksLock.unlock() }
```
tanderson-ld, line 308: defer is not always without cost.  Do you need defer here?

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -281,10 +281,34 @@ public class LDClient {
...
+    /// way by a series running on another.
+    private var storedHooks: [Hook]
+    private let hooksLock = NSLock()
+
+    /// A snapshot of the hooks to run. Every series reads this once into a local and works from that, so the hooks a
+    /// series ends with are the hooks it began with: were it read again mid-series, a hook registered in between would
+    /// be given an "after" stage for a series whose "before" stage it was never in.
+    var hooks: [Hook] {
+        hooksLock.lock()
+        defer { hooksLock.unlock() }
```
tanderson-ld, line 295: defer is not always without cost. Do you need defer here?

LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -745,6 +770,26 @@ public class LDClient {
...
+
+     This registers the plugin with this client only. In a multi-environment configuration each environment has its own
+     client, so registering with every environment means calling this on each of them.
+
+     - parameter plugin: The plugin to register.
+     */
+    public func registerPlugin(_ plugin: Plugin) {
+        let pluginHooks = plugin.getHooks(metadata: environmentMetadata)
+        plugin.register(client: self, metadata: environmentMetadata)
+        addHooks(pluginHooks)
```
tanderson-ld: I think addHooks should be before register.

```
While hooks could be added during register, if the SDK supports runtime hook registration, this could result in a plugin missing hook invocations. This is because other plugins could call track, identify, or variation methods during registration. Getting a list of hooks before
  registration ensures that the plugin cannot miss any operations.
```

LaunchDarkly/LaunchDarkly/LDClient.swift
abelonogov-ld: LDClient is the class and defer would be negligible compare to lock/unlock itself

LaunchDarkly/LaunchDarkly/LDClient.swift
abelonogov-ld: LDClient is the class and defer would be negligible compare to lock/unlock itself

abelonogov-ld: defer is considered good pattern for locking

tanderson-ld (approved):

PR #525, opened 2026-08-25
fix: activate a plugin's hooks only after it has registered

abelonogov-ld:
## Summary

Follow-up to #522. A plugin's hooks now go live only once every plugin has registered, and this applies to plugins in `LDConfig.plugins` and to `registerPlugin` alike, so a plugin behaves the same however it was registered.

Previously `registerPlugin` activated a plugin's hooks before calling `register`, which meant a plugin's own hooks observed the evaluations and identify calls its `register` made. Activating them last also means no plugin's hooks observe *another* plugin's `register`, rather than that depending on where the plugin sat in the list.

Both paths now share one `registerPlugins` method, so the two cannot drift apart again.

### Keeping plugin hooks in the initial identify

Activating hooks after `register` is only safe if registration still happens before the first identify series opens. On iOS it did not: the series began inside `init`, while `plugin.register` ran afterwards in `start`. Left alone, configured plugin hooks would have missed the initial identify entirely, since `afterIdentify` replays the snapshot taken at `init`.

So the tail of `init` moved into `startIdentifyAndGoOnline`, and `start` now:

1. constructs the instance,
2. publishes it to `LDClient.instances` (so a plugin calling `LDClient.get()` from `register` still finds it),
3. registers the configured plugins,
4. opens the first identify series and goes online.

`collectHooks` is gone, since `init` now seeds only the configuration's own hooks. `executeBeforeIdentifyHooks` takes a `methodName` so the startup path can reuse it rather than duplicating the series setup.

## Test plan

- [x] Full suite passes (642 tests)
- [x] `testPluginHooksFireDuringInit` still passes, confirming configured plugin hooks continue to take part in the initial identify
- [x] `testRegisterPluginDoesNotRunTheRegisteringPluginsOwnHooks` covers the new ordering for the runtime path
- [x] New `testConfiguredPluginHooksDoNotObserveAnotherPluginsRegister` covers the cross-plugin case

---

> [!NOTE]
> **Overview**
> Plugin hooks now activate **after** each plugin’s `register` returns, and only once **all** plugins in a batch have registered. **`registerPlugin`** and **`LDConfig.plugins`** both use shared **`registerPlugins`**, so runtime and config-time registration behave the same: a plugin’s hooks no longer see its own (or another plugin’s) flag evaluations or identify calls made inside `register`.
> 
> Startup was reordered so configured plugin hooks still participate in the **initial identify**. The first identify / go-online work moved from **`init`** into **`startIdentifyAndGoOnline`**, and **`start`** now publishes the client to **`LDClient.instances`**, runs **`registerPlugins`**, then opens that series. **`init`** seeds **`storedHooks`** from config hooks only ( **`collectHooks`** removed). **`executeBeforeIdentifyHooks`** accepts a **`methodName`** so startup can label the series `"init"`.
> 
> Tests were updated for the new hook timing and add coverage that one configured plugin’s hooks do not run during another plugin’s **`register`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4002fcebac09af1035e294fd3c8d5b15848cb637. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/LaunchDarkly/LaunchDarkly/LDClient.swift
+++ b/LaunchDarkly/LaunchDarkly/LDClient.swift
@@ -774,9 +774,9 @@ public class LDClient {
      Registers a single plugin with this client after the client has been configured and started. To register plugins
      before the client starts, set `LDConfig.plugins` instead.
 
-     The plugin's hooks begin running before `Plugin.register` is called, as they do for a plugin registered at
-     configuration time, so a plugin's own hooks observe the flag evaluations and identify calls that its `register`
-     makes. A plugin whose `register` fails keeps contributing its hooks, again matching configuration time.
+     The plugin's hooks begin running only once `Plugin.register` has returned, so a plugin's own hooks do not observe
+     the flag evaluations or identify calls that its `register` makes. This is also how the plugins in
+     `LDConfig.plugins` are registered, by this same method, so a plugin behaves the same however it was registered.
 
      This registers the plugin with this client only. In a multi-environment configuration each environment has its own
      client, so registering with every environment means calling this on each of them.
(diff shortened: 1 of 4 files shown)


tanderson-ld:
I don't understand the need for this change?  I think this violates the intent of getHooks existing.

tanderson-ld (commented):
I don't understand the need for this.  Consider slack messaging to help me understand.  See my other comment https://github.com/launchdarkly/ios-client-sdk/pull/525#issuecomment-5426383879

PR #530, opened 2026-09-19
perf(events): encode events with a hand-written JSON writer (tier 1)

abelonogov-ld:
Tier 1 of the mobile event durability work. This tier is the event buffer and nothing else: no persistence, no bounded flush. Those are tiers 2 and 3, which stack on this branch.

## What this changes

The event path is the one place the SDK encodes often enough for `Codable`'s reflection to show up in a profile. This writes the wire form directly into a byte buffer instead — the same approach the Android SDK takes — and caches the encoded context, because a run of evaluations is nearly always the same context over and over.

Getting there needed one change in the model. A feature event used to redact its context by deep-copying `LDContext` and setting a flag on the copy. That cost an allocation per event, and it defeated the buffer-identity comparison that makes a context-keyed cache affordable: `Dictionary` and `Array` equality check buffer identity before contents, so two contexts sharing storage compare equal in nanoseconds instead of by walking every attribute. Redaction is really a property of the *event* rather than of the context — one context redacts on a feature event and does not on the debug event beside it — so it now lives on `Event`, and the context is held as the caller gave it.

The summarizer keys its trackers by context rather than by a digest of one, for the same reason.

## Why this is safe

`JSONEncoder` is kept as the oracle. `EventJSONWriterTests` asserts the two encoders produce identical bytes for every event shape the SDK emits, and where they disagree `Codable` is right. The reporter still exposes `.codable` so that comparison is a real switch rather than a claim.

## Notes for review

- `EventReporter`'s encoder is built once rather than per delivery, and one `JSONWriter` covers a whole run so its buffer is reused across the events in it.
- `LDContext` gains a small internal seam (`redactingAnonymousAttributes()`, `redactionDecision(...)`) so the hand-written writer reuses the existing redaction logic rather than reimplementing it. The two encoders must differ in how bytes are produced, not in what gets redacted.

Spec: [Event Durability](https://github.com/launchdarkly/ios-client-sdk) §3.1, §5.

Made with [Cursor](https://cursor.com)


LaunchDarkly/LaunchDarkly/LDClient.swift
```diff
@@ -293,7 +293,7 @@ public class LDClient {
     var hooks: [Hook] {
         hooksLock.lock()
         defer { hooksLock.unlock() }
-        return storedHooks
+        return Array(storedHooks)
```
tanderson-ld: What's up with this?

LaunchDarkly/LaunchDarkly/ServiceObjects/JSONWriter.swift
tanderson-ld: Aren't there existing JSON writer libraries that can be used to get the performance of direct writing, but that don't require us to re-implement JSON syntax handling?

LaunchDarkly/LaunchDarkly/ServiceObjects/JSONWriter.swift
```diff
@@ -0,0 +1,249 @@
...
+            }
+            appendEscape(buffer[index])
+            runStart = index + 1
+        }
+        if runStart < buffer.count {
+            bytes.append(contentsOf: buffer[runStart...])
+        }
+    }
+
+    private func appendEscape(_ byte: UInt8) {
```
tanderson-ld: What spec is this following?

LaunchDarkly/LaunchDarkly/ServiceObjects/JSONWriter.swift
```diff
@@ -0,0 +1,249 @@
...
+        if magnitude == 0 {
+            bytes.append(UInt8(ascii: "0"))
+            return
+        }
+
+        let start = bytes.count
+        while magnitude > 0 {
+            bytes.append(UInt8(ascii: "0") + UInt8(magnitude % 10))
+            magnitude /= 10
+        }
```
tanderson-ld: Is this more performant than a community swift library for json writing?  This feels like re-inventing the wheel.

And to be more clear, I'm referring to a library that lets you do json object writing piece by piece and not serializing whole classes via annotations/reflection etc...

LaunchDarkly/LaunchDarkly/Models/Context/LDContext.swift
```diff
@@ -48,15 +48,20 @@ public struct LDContext: Encodable, Equatable {
         self.canonicalizedKey = canonicalizedKey
     }
 
-    internal init(copyFrom: LDContext) {
```
tanderson-ld: Unused?

LaunchDarkly/LaunchDarkly/Models/Context/LDContext.swift
```diff
@@ -48,15 +48,20 @@ public struct LDContext: Encodable, Equatable {
...
+    /// A copy of this context that redacts anonymous attributes when it is encoded.
+    ///
+    /// `LDContext` is a value type, so this is an ordinary struct copy: the attribute dictionary and the sub-context
+    /// array are shared until one side mutates, and nothing is rebuilt. **The sharing matters for more than the
+    /// allocation it saves.** `Dictionary` and `Array` equality compare buffer identity before contents, so two
+    /// contexts related by this copy compare equal in single-digit nanoseconds instead of by walking every attribute --
+    /// which is what makes a context-keyed cache affordable.
+    ///
+    /// Only the top-level flag is ever read. `encode(to:)` passes the parent's value down to every sub-context and
+    /// ignores whatever theirs holds, so the sub-contexts do not need rewriting here.
```
tanderson-ld: Why does this comment talk about `encode(to:) passes the parent's value down to every sub-context` ?  That seems not relevant to this function and is relevant in the encode commenting.

LaunchDarkly/LaunchDarkly/Models/Context/LDContextJSONWriter.swift
```diff
@@ -0,0 +1,157 @@
+import Foundation
+
+/// The hand-written counterpart to `LDContext.encode(to:)`, for the encoding experiment behind `EventJSONWriter`.
+///
+/// The field set, the omissions, and the redaction rules are deliberately identical to the `Codable` path; where the
+/// two disagree, the `Codable` path is right and this is wrong. `EventJSONWriterTests` asserts they agree.
```
tanderson-ld: This AI comment needs to be improved.  It shouldn't talk about development history and such.

LaunchDarkly/LaunchDarkly/Models/Context/LDContextJSONWriter.swift
tanderson-ld: I think there are a good number of context serialization contract tests in the sdk test harness.  It may be worth pointing AI at the sdk-test-harness to find those cases to get the same amount or more of case coverage.

LaunchDarkly/LaunchDarkly/ServiceObjects/JSONWriter.swift
```diff
@@ -0,0 +1,249 @@
...
+
+    func write(_ value: Int64) {
+        separate()
+        writeInteger(value)
+        needsSeparator = true
+    }
+
+    /// Integral values are written without a fractional part, which is what `JSONEncoder` does and what the service
+    /// already receives. Anything else takes Swift's shortest round-trip description, which is valid JSON.
+    func write(_ value: Double) {
```
tanderson-ld: Should we really be implementing this when community libs have this? Same for the other primitive types.

LaunchDarkly/LaunchDarkly/Models/Context/LDContextJSONWriter.swift
tanderson-ld: Attribute references tends to be a tricky aspect of this sort of logic.  How are you confirming you have proper attribute reference handling?

LaunchDarkly/LaunchDarkly/Models/Context/LDContext.swift
abelonogov-ld: replaced by real struct copy, manual copying is slow

LaunchDarkly/LaunchDarkly/ServiceObjects/EventReporter.swift
```diff
@@ -51,10 +51,27 @@ class EventReporter: EventReporting {
...
+    /// The reflective encoder, built once and then only read.
+    ///
+    /// `JSONEncoder` is `@unchecked Sendable` and constructs a fresh internal encoder for each `encode` call, so one
+    /// instance can be shared — but only while nothing mutates it, which is why `userInfo` is set here rather than per
+    /// call. What it holds cannot go stale: `config` is a `let` on the service, so the privacy settings the encoding
+    /// depends on are fixed for as long as this reporter exists.
+    private let encoder: JSONEncoder
+
+    let encoding: Encoding
+    private let handWrittenEncoder: EventJSONWriter
```
tanderson-ld:
```suggestion
    private let jsonWriter: EventJSONWriter
```

LaunchDarkly/LaunchDarkly/ServiceObjects/JSONWriter.swift
```diff
@@ -0,0 +1,249 @@
...
+        separate()
+        if value.isFinite, value.rounded() == value, let exact = Int64(exactly: value.rounded()) {
+            writeInteger(exact)
+        } else if value.isFinite {
+            bytes.append(contentsOf: String(value).utf8)
+        } else {
+            bytes.append(contentsOf: JSONWriter.nullBytes)
```
tanderson-ld: NaN/±Inf silently become `null` here, where the `Codable` path throws (which today drops the whole event batch via `try? encode`). The parity tests can't cover this case since `JSONEncoder` refuses these values — worth an explicit decision either way.

LaunchDarkly/LaunchDarkly/Models/Context/LDContextJSONWriter.swift
```diff
@@ -0,0 +1,157 @@
...
+                                   key: name,
+                                   value: value,
+                                   redaction: &redaction)
+        }
+        let redactedAttributes = redaction.redactedAttributes
+
+        // Matches `Meta.isEmpty` and `Meta.encode`: `_meta` is written whenever either list is non-empty, but
+        // `privateAttributes` is only included when the caller asked for it, which the event path never does. A context
+        // with private attributes and nothing redacted therefore writes an empty `_meta`, as it does today.
+        if !privateAttributes.isEmpty || !redactedAttributes.isEmpty {
```
tanderson-ld: `trySetValue` allows a custom attribute literally named `_meta`, which the loop above already wrote — this block then emits a second `_meta` key in the same object (RFC 8259 says receiver behavior is unpredictable). The `Codable` path collapses to one key, last write wins.

LaunchDarkly/LaunchDarkly/ServiceObjects/ContextEncodingCache.swift
```diff
@@ -0,0 +1,110 @@
...
+    private static func slot(_ redactAnonymous: Bool) -> Int {
+        redactAnonymous ? 1 : 0
+    }
+
+    /// `Reference` compares and hashes its *parsed components* and ignores the raw string it was built from, so
+    /// `Reference("name") == Reference("/name")`. Redaction writes `raw()` into `_meta.redactedAttributes`, so two
+    /// contexts that differ only in how a private attribute was spelled are `==` yet encode differently. `==` alone is
+    /// therefore not a sound key, and the spellings are compared separately.
+    ///
+    /// Contexts with no private attributes -- the common case -- settle this without allocating.
```
tanderson-ld: Isn't there a way to do this without allocating at all?

tanderson-ld: Concretely: by the time this runs, `entry.context == context` has already passed, and `==` covers `privateAttributes` — so the two sets are equal as *parsed* references, and the only open question is spelling drift between equal members. `Set.firstIndex(of:)` returns the *stored* member, so the check can probe the cached set directly:

```swift
for ref in context.privateAttributes {
    guard let i = entry.context.privateAttributes.firstIndex(of: ref),
          entry.context.privateAttributes[i].raw() == ref.raw()
    else { return false }
}
// then recurse pairwise over zip(entry.context.contexts, context.contexts)
```

Hashed probe plus stored-string compare: no arrays, no `sorted()`, and `Entry.spellings` / `spellings(of:)` / the empty-set fast path all go away — `store` stops allocating too. This leans on `Reference` hashing by parsed components rather than `rawPath`, which its `Hashable` contract already requires given its `==`.

tanderson-ld (commented):
Ran the multi-agent review battery over this branch (~195k differential-fuzz comparisons against the `JSONEncoder` oracle, plus TSan stress runs). Redaction parity, cache slot correctness, escaping, and flush concurrency all came back clean. High/Medium findings below, summarized — I have failing proof tests for the two Highs and a longer Low/informational list if useful.

**High**

1. **Duplicate `_meta` member.** An attribute literally named `_meta` is accepted by `trySetValue` (java-core rejects it as reserved) and emitted by the attribute loop, then the redaction block emits the SDK's `_meta` again — two members in one object. `Codable` masked this (repeat keys replace); first-wins parsers now read the app's value and drop `redactedAttributes`, and an app-supplied `_meta` that never left the device now ships. Suggest rejecting `_meta` in the builder + adding it to the test corpus.
2. **Summarizer key hashes only the canonical key** (`ContextSummarizer.ContextKey`), so same-key/different-attribute contexts pile into one probe chain of deep `LDContext ==` walks — synchronous on the `variation()` caller's thread, growing until a successful flush. Measured 57x per-insert at N=2000 (linear growth, crossover ≈ N=30) via the mundane re-identify-with-a-changing-attribute pattern. Count-based hash discriminators won't fix it; needs attribute values in the hash or a bounded/digest fallback.
3. **The `fix crash` commit (`e70cb42a`) looks like three inert edits** — the `Data`/`Array` aliasing its comments describe doesn't reproduce, the `LDClient.hooks` `Array()` copy fixes nothing (and adds a per-evaluation allocation on the hottest path), and the `takeStages` change is identical semantics. If so, the actual crash is still unexplained — can you share the crash log and whether it reproduces with the `JSONWriter` change alone?

**Medium**

4. **Cache can serve wrong `_meta.redactedAttributes` spellings for multi-contexts** — `spellings(of:)` flattens+sorts across sub-contexts, so per-part spelling permutations collide. No value leak; fix is sorting per context and concatenating in `contexts` order.
5. **Non-finite doubles become `null`; `"metricValue": null` violates the v4.0 custom-event schema** (and diverges from Android, which drops). Omit the key for non-finite metrics; have `unserializableEventSpec` assert payload bytes.
6. **Integral doubles in `[1e17, 9.22e18]` diverge from `JSONEncoder` even after canonicalization** — adding `1e17` to `testNumberFormatting` fails today. Widen the fast-path guard to `magnitude < 1e17` or document it (and restate the "identical bytes" claim as JSON-equality; `/`-escaping and `-0.0` also differ at byte level).
7. **No runtime escape hatch back to `.codable`** — the default wire path has no config switch, so the field remedy for any divergence is a release.
8. **The shipping configuration (`.handWrittenCachingContext` + anonymous context + privacy config) is never exercised at reporter level** — the reporter-level encoder comparison uses `.codable` vs `.handWritten`, and reporter specs use a non-anonymous stub. One mixed-batch reporter test on the default encoding would close it.

LaunchDarkly/LaunchDarkly/ServiceObjects/ContextEncodingCache.swift
```diff
@@ -0,0 +1,110 @@
...
+        let spellings: [String]
+        let encoded: [UInt8]
+    }
+
+    private let lock = UnfairLock()
+    private var entries: [Entry?] = [nil, nil]
+
+    /// Counted only so the experiment can report a hit rate rather than assume one.
+    private(set) var hits = 0
+    private(set) var misses = 0
```
tanderson-ld: Were these just for debugging inspection?

